PR #2, opened 2017-08-16
WIP PR to defecting master

scripttease:
Kindle-gen queries the print-sent API (CAPI Preview) and creates articles from the response for conversion to NITF format.

The Manifest class is used to create a contents manifest for the kindle files.

The BookSection and BookSectionPage classes will be used to sort the articles according to section and page number.

The requirements for the NITFs and kindle files can be found (here)[https://images-na.ssl-images-amazon.com/images/G/01/kindle-publication/feedGuide-new/KPPUserGuide._V181169266_.html#FeedSpecifications].


build.sbt
```diff
@@ -6,7 +6,7 @@ description:= "Converting content to NITF format"
 
 version := "1.0"
 
-scalaVersion := "2.12.1"
+scalaVersion := "2.11.5"
```
tsop14: Why did you have to use an older version?

annebyrne: I think this is because we haven't yet upgraded to `2.12` on our client, so Alice was having some dependancy problems

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -48,3 +61,96 @@ object TestIt {
...
+    response.results
+  }
+
+  def resultToArticles(response: Seq[com.gu.contentapi.client.model.v1.Content]): Seq[Article] = {
+    response.flatMap(responseContent =>
+      responseContent.fields.map(fields =>
+        makeArticle(fields, responseContent)))
+  }
+
+  private def makeArticle(fields: com.gu.contentapi.client.model.v1.ContentFields, responseContent: com.gu.contentapi.client.model.v1.Content): Article =
```
annebyrne: Could this not be a overridden apply method in the `Article` case class?

build.sbt
```diff
@@ -15,19 +15,35 @@ scalacOptions ++= Seq(
   "-Ywarn-dead-code"
 )
 
+/* deps for aws lambda */
 libraryDependencies ++= Seq(
-  "com.amazonaws" % "aws-lambda-java-core" % "1.1.0"
+  "com.amazonaws" % "aws-lambda-java-core" % "1.1.0",
+  "com.gu" %% "content-api-client" % "11.22",
+  "org.typelevel" %% "cats" % "0.9.0",
+
```
annebyrne: whitespace

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -0,0 +1,59 @@
...
+    releaseDate: CapiDateTime,
+    pubDate: CapiDateTime,
+    byline: String,
+    articleAbstract: String, // standfirst is used
+    content: String
+) {
+  def toNitf: String = {
+    val formattedIssueDate = formatter.print(isoFormatter.parseDateTime(capiIsoDateTimeToString(issueDate)))
+    val formattedReleaseDate = formatter.print(isoFormatter.parseDateTime(capiIsoDateTimeToString(releaseDate)))
+    val formattedPubDate = formatter.print(isoFormatter.parseDateTime(capiIsoDateTimeToString(pubDate)))
```
annebyrne: could we abstract out `formatter.print(isoFormatter.parseDateTime(capiIsoDateTimeToString(_)))` ? we need it for all 3 `val`s

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -0,0 +1,59 @@
...
+       |<body.end />
+       |</body>
+       |</nitf>""".stripMargin
+  }
+
+  // formatter for use with parseDateTime to convert an isoDateTime: String to isoDateTime: DateTime
+  private def isoFormatter = ISODateTimeFormat.dateTime()
+  // formatter for use with parseDateTime or print to convert a DateTime: Long to DateTime: DateTime
+  private def formatter = DateTimeFormat.forPattern("yyyyMMdd")
+  private def capiIsoDateTimeToString(dt: CapiDateTime): String = dt.iso8601
```
annebyrne: maybe a more descriptive param name?

src/main/scala/com/gu/kindlegen/BookSectionPage.scala
```diff
@@ -0,0 +1,7 @@
+package com.gu.kindlegen
```
annebyrne: Given this and `BookSection` are both so small, could we combine them in the same file? Keeping them seperate is a nice touch and normally what I would advocate but might be helpful to see them side by side in the same file.

If you anticipate writing v long methods for each though I wouldn't bother with combining them in the same file.

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -48,3 +61,96 @@ object TestIt {
...
   }
 }
+
+object Querier {
+
+  class PrintSentContentClient(override val apiKey: String) extends GuardianContentClient(apiKey) {
+    // This overrides the val for targetUrl from "http://content.guardianapis.com"
+    // the `preview.` is required in order to access the print-sent api.
+    override val targetUrl = "https://preview.content.guardianapis.com/content/print-sent"
+    // TODO: Find out if capi Auth or the trailing "&user-tier=internal" that is used in the kindle-previewer is required
```
annebyrne: We shouldn't probably have an internal API key / param checked in to a project.

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -48,3 +61,96 @@ object TestIt {
...
+
+  }
+
+  def readApiKey = {
```
annebyrne: I might be making this up (I probably have totally imagined it :joy: ) but there might be a simpler way of doing this in one of the other projects. I'd check Floodgate and the Shuttlerun lambda within the `content-api` repo

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -48,3 +61,96 @@ object TestIt {
...
+      .orderBy("newest")
+      //      .pageSize(fetchPageSize) // This is meaningless as not defined in previewer?
+      .fromDate(editionDateStart)
+      .toDate(editionDateEnd)
+      .useDate("newspaper-edition")
+      //      .page(pageNum)
+      .showFields("newspaper-page-number, headline,newspaper-edition-date,byline,standfirst,body") // TODO: what fields are required? main? content?
+      .showTags("newspaper-book-section, newspaper-book")
+      .showElements("image")
+    // TODO: Add error handling with Try for failed request.
```
annebyrne: Could you have the method return an `Either[Error, Success]` or an `Option` after the query? That is sometimes easier to parse / work with than a `Try`

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -48,3 +61,96 @@ object TestIt {
...
+      pubDate = fields.newspaperEditionDate.get,
+      byline = fields.byline.getOrElse(""),
+      articleAbstract = fields.standfirst.getOrElse(""),
+      content = fields.body.getOrElse("")
+    // Note `body` used here which includes html tags. (`bodyText` strips tags)
+    )
+
+  // Pass in Querier.resultToArticles(getPrintSentResponse)
+  def toSectionHeading(articles: Seq[Article]): Seq[SectionHeading] = {
+    articles.map(x =>
```
annebyrne: How about `article` rather than `x`?

src/test/scala/com/gu/kindlegen/ArticleSuite.scala
```diff
@@ -0,0 +1,59 @@
...
+import com.github.nscala_time.time.Imports._
+import com.gu.contentapi.client.model.v1.CapiDateTime
+import com.gu.contentapi.client.utils._
+
+@RunWith(classOf[JUnitRunner])
+class ArticleSuite extends FunSuite {
+  import Article._
+
+  test("Article.toNitf") {
+    val expectedOutput = """
```
annebyrne: Can we put expected output in files in a seperate `test/resources` folder so they won't clutter the tests?

annebyrne:
Great work so far! 💯

build.sbt
scripttease: Yep - in fact it won't work in 2.12 - this might be a project that I would be interested in picking up if it isn't too large? I've become v familiar with the client now because of the kindle-gen work. Who is it best to discuss this with?

src/main/scala/com/gu/kindlegen/Article.scala
scripttease: YEP! :)

src/main/scala/com/gu/kindlegen/BookSectionPage.scala
scripttease: Probably both will have long methods but will keep this in mind - agree that it couldve been more sensible to start in same file and move if/when necessary (as long as one does move them!)

build.sbt
annebyrne: Just check the Trello card and pick any up that you're interested in! ☺️ 
https://trello.com/c/QTUArrS9/80-update-to-scala-212x

src/main/scala/com/gu/kindlegen/Lambda.scala
scripttease: I thought that because it is in `~/.gu/kindle-gen.conf` it isn't checked in to VC? Not seeing where it is, help?

annebyrne: Ah yes but what I meant was 
` // TODO: Find out if capi Auth or the trailing "&user-tier=internal" that is used in the kindle-previewer is required` * "&user-tier=internal"* will now be included in version history

build.sbt
```diff
@@ -15,19 +15,35 @@ scalacOptions ++= Seq(
...
+
+  "org.scalactic" %% "scalactic" % "3.0.1",
+  "org.scalatest" %% "scalatest" % "3.0.1" % "test",
```
tsop14: Included twice

src/main/scala/com/gu/kindlegen/Lambda.scala
scripttease: Argh ok yes...  :/ Eeeek. Can we do anything about this? oooops...

annebyrne: Let's just remove the comment for now. In a best case scenario we would use a tool (think @tsop14 would be more knowledgeable about this than me) that would prune repo of any mention.

src/main/scala/com/gu/kindlegen/TestArticle.scala
```diff
@@ -0,0 +1,126 @@
...
+import com.gu.contentapi.client.model.v1._
+import com.gu.contentapi.client.model.v1.ContentType
+import com.gu.contentapi.client.model.v1.TagType.NewspaperBookSection
+import com.gu.contentapi.client.model.v1.Content
+import com.gu.contentapi.client.model.v1.ContentFields
+import com.gu.contentapi.client.utils.CapiModelEnrichment
+
+/**
+ * Created by alice_dee on 23/08/2017.
+ */
```
mchv: this can be remove as we have history in git

PR #3, opened 2017-08-25
PR to WIP master - Tidies up classes and methods resulting from review.

scripttease:
This PR will replace the previous WIP PR, containing *some* of the fixes:
(Note, not rebased *yet* to keep the commits small and easier to refer to)
- removes url to external `.conf` file (for now, pre deployment as Lambda)
- refactors the Querier object moving makeArticle method to an apply method on the Article case class
- refactors the toNITF method so that it is an apply method on an ArticleNITF class
- adds a TestContent class to make it easy to generate sample content for testing without hitting the external print-sent API.

Kindle-gen queries the print-sent API (CAPI Preview) and creates articles from the response for conversion to NITF format.

The Manifest class is used to create a contents manifest for the kindle files.

The BookSection and BookSectionPage classes will be used to sort the articles according to section and page number.

The requirements for the NITFs and kindle files can be found (here)[https://images-na.ssl-images-amazon.com/images/G/01/kindle-publication/feedGuide-new/KPPUserGuide.V181169266.html#FeedSpecifications].


build.sbt
```diff
@@ -6,7 +6,7 @@ description:= "Converting content to NITF format"
 
 version := "1.0"
 
-scalaVersion := "2.12.1"
+scalaVersion := "2.11.5"
```
mchv: `2.11.11`

build.sbt
```diff
@@ -15,19 +15,35 @@ scalacOptions ++= Seq(
   "-Ywarn-dead-code"
 )
 
+/* deps for aws lambda */
 libraryDependencies ++= Seq(
-  "com.amazonaws" % "aws-lambda-java-core" % "1.1.0"
+  "com.amazonaws" % "aws-lambda-java-core" % "1.1.0",
+  "com.gu" %% "content-api-client" % "11.22",
+  "org.typelevel" %% "cats" % "0.9.0",
```
mchv: I don't think you need this dependency at this point

build.sbt
```diff
@@ -15,19 +15,35 @@ scalacOptions ++= Seq(
...
+  "org.scalactic" %% "scalactic" % "3.0.1",
+  "org.scalatest" %% "scalatest" % "3.0.1" % "test",
+  "junit" % "junit" % "4.10" % "test",
+  "com.novocode" % "junit-interface" % "0.11" % "test"
 )
 
+/* deps required to use junit test and test watch */
 libraryDependencies += "org.scalactic" %% "scalactic" % "3.0.1"
 libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % "test"
 libraryDependencies += "junit" % "junit" % "4.10" % "test"
```
mchv: I am not sure why you need `junit`we usually run all the tests without creating suites

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -48,3 +61,175 @@ object TestIt {
...
+
+  // Pass in Querier.resultToArticles(getPrintSentResponse)
+  def toManifest(articles: Seq[Article], buildDate: DateTime = DateTime.now): SectionManifest = {
+    SectionManifest(
+      publicationDate = articles.head.pubDate,
+      buildDate = buildDate,
+      sections = Querier.toSectionHeading(articles)
+    )
+  }
+}
```
mchv: The object `Querier` should be in its own scala file

src/main/scala/com/gu/kindlegen/TestContent.scala
```diff
@@ -0,0 +1,144 @@
...
+    webPublicationDate = None,
+    debug = None,
+    blocks = None,
+    isHosted = false,
+    webTitle = "",
+    sectionId = None,
+    elements = Option(Nil),
+    isExpired = None
+  )
+}
```
mchv: that should be in `test` repository not in `main`

src/test/scala/com/gu/kindlegen/TestArticleSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+
+  test("TestContent as factory") {
+    val ta = TestContent("", "", 1, "", "", CapiModelEnrichment.RichJodaDateTime(formatter.parseDateTime("20170724")).toCapiDateTime, "", "", "")
+    assert(ta.toContent.fields.flatMap(_.headline) === Some(""))
+    val ta2 = ta.copy(testArticleTitle = "new title")
+    assert(ta2.toContent.fields.get.headline === Some("new title"))
+    // add default args
+  }
+
+}
```
mchv: Rather than creating `Suite` you create a class ending by `Spec` and it will be discovered and run automatically when launching `test` see for instance https://github.com/guardian/facebook-ia-rss/blob/master/test/com/gu/facebook/AdSpec.scala

src/main/scala/com/gu/kindlegen/TestContent.scala
scripttease: Corrected in commit #88d85ae

src/main/scala/com/gu/kindlegen/Lambda.scala
scripttease: resolved in commit (2d0745ba53420874fc7d098224e2b17166660e9e)[https://github.com/guardian/kindle-gen/commit/2d0745ba53420874fc7d098224e2b17166660e9e]

src/test/scala/com/gu/kindlegen/TestArticleSuite.scala
scripttease: Hi Mariot, I have refactored QuerierSpec to use FlatSpec instead of JUnit. 
commit b7368fbd463647e6f3de0ac296db70e1db674de1

I only used JUnit as it feels to me very slightly faster to run and to write (it still works with sbt ~test or sbt test) and I use unit tests with the test watcher `sbt ~test` all the time as I write code, it is my feedback loop. I find the syntax of JUnit a bit faster and clearer for simple unit tests. I have changed it in this case, I found it is very similar and I can refactor all the tests to use this if necessary. Because of the libraries and deps I am using, any `sbt test` command in teamCity or a github hook should still trigger all the tests whichever test library I use and it will also pick up all the test files whether they are called 'Suite' or 'Spec'. From now on I am happy to switch to Spec and Scalatest to be consistent with the rest of the team.

A lot of the code I have seen so far has only integration tests so I just picked my own unit testing library 🌞

build.sbt
scripttease: See answer below (I am happy to change all my test files, I have changed one)

PR #6, opened 2017-09-08
Image field to article

scripttease:
A capi Response is converted to a  Seq[Article]. Each Article contains the attribute Option(imageUrl), because there may or may not be an imageURL that corresponds to an image that meets the filtering requirements for NITF articles.

fetchImageData retrieves the images for the articles that DO have a suitable image, and is now a Future so that when pagination is added, this method can be called on, for example, the first page of capi Articles, while successive pages are being retrieved.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/Article.scala
+++ b/src/main/scala/com/gu/kindlegen/Article.scala
@@ -32,10 +32,10 @@ object Article {
     val elements = content.elements
     val mainImage = elements.flatMap(_.find(isMainImage))
     val imageAsset = mainImage.flatMap(_.assets.find(isCorrectSizedAsset))
+    // keep imageUrl as an option otherwise later you will try to download form an empty string
     val imageUrl = imageAsset.flatMap(_.typeData).flatMap(_.secureFile)
     imageUrl
   }
-  // keep this as an option otherwise later you will try to download form an empty string
   // TODO: get rid of the gets
   def apply(contentWithIndex: (Content, Int)): Article = {
     val content = contentWithIndex._1
--- a/src/main/scala/com/gu/kindlegen/Article.scala
+++ b/src/main/scala/com/gu/kindlegen/Article.scala
@@ -35,7 +35,7 @@ object Article {
     imageUrl
   }
   // keep this as an option otherwise later you will try to download form an empty string
-
+  // TODO: get rid of the gets
   def apply(content: Content): Article = new Article(
     newspaperBookSection = content.tags.find(_.`type` == NewspaperBookSection).get.id,
     sectionName = content.tags.find(_.`type` == NewspaperBookSection).get.webTitle,
--- a/src/main/scala/com/gu/kindlegen/Article.scala
+++ b/src/main/scala/com/gu/kindlegen/Article.scala
@@ -1,5 +1,8 @@
 package com.gu.kindlegen
 
+import com.gu.contentapi.client.model.v1.Asset
+import com.gu.contentapi.client.model.v1.Element
+import com.gu.contentapi.client.model.v1.ElementType.Image
 import com.gu.contentapi.client.model.v1.TagType.NewspaperBookSection
 import com.gu.contentapi.client.model.v1._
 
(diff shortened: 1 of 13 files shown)


src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -10,26 +13,46 @@ case class Article(
...
+      a.typeData.flatMap(_.width) == Some(500)
+
+    val elements = content.elements
+    val mainImage = elements.flatMap(_.find(isMainImage))
+    val imageAsset = mainImage.flatMap(_.assets.find(isCorrectSizedAsset))
+    // keep imageUrl as an option otherwise later you will try to download form an empty string
+    val imageUrl = imageAsset.flatMap(_.typeData).flatMap(_.secureFile)
+    imageUrl
+  }
+  // TODO: get rid of the gets
```
annebyrne: I think its fine to use `get`, however it would be preferable to use `getOrElse`.

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -10,26 +13,46 @@ case class Article(
...
-    pubDate = content.fields.flatMap(_.newspaperEditionDate).get,
-    byline = content.fields.flatMap(_.byline).getOrElse(""),
-    articleAbstract = content.fields.flatMap(_.standfirst).getOrElse(""),
-    content = content.fields.flatMap(_.body).getOrElse("")
-  )
+  def getImageUrl(content: Content): Option[String] = {
+    def isMainImage(e: Element): Boolean =
+      e.`type` == Image && e.relation == "main"
+    def isCorrectSizedAsset(a: Asset): Boolean =
+      a.typeData.flatMap(_.width) == Some(500)
```
regiskuckaertz: I think `....exists(_ == 500)` is more idiomatic?

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -30,34 +34,66 @@ object Querier {
...
-      Article(responseContent))
+  // TODO: handle the possibility of there being no content in the getPrintSentResponse method above
+  def responseToArticles(response: Seq[Content]): Seq[Article] = {
+    val sortedContent = Querier.sortContentByPageAndSection(response)
+    val contentWithIndex = sortedContent.view.zipWithIndex.toList
+    contentWithIndex.map(Article.apply)
+  }
+
+  def sortContentByPageAndSection(response: Seq[Content]): Seq[Content] = {
+    response.sortBy(content => (content.fields.flatMap(_.newspaperPageNumber), content.tags.find(_.`type` == NewspaperBookSection).get.id))
```
annebyrne: could this be split over multiple lines to be more legible?

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -10,26 +13,46 @@ case class Article(
...
+
+    val elements = content.elements
+    val mainImage = elements.flatMap(_.find(isMainImage))
+    val imageAsset = mainImage.flatMap(_.assets.find(isCorrectSizedAsset))
+    // keep imageUrl as an option otherwise later you will try to download form an empty string
+    val imageUrl = imageAsset.flatMap(_.typeData).flatMap(_.secureFile)
+    imageUrl
```
regiskuckaertz: It is for you to decide, but I often find `flatMap` confusing to follow:
```
for {
  elements <- elements
  mainImage <- elements.find(isMainImage)
  imageAsset <- mainImage.assets.find(isCorrectSizedAsset)
  typedData <- imageAsset.typedData
  imageUrl <- typedData.secureFile
} yield imageUrl
```

annebyrne: +1. When you're using multiple maps & flatMaps, I find it easier to break over multiple lines or use a `for...yield`

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -30,34 +34,66 @@ object Querier {
...
+            case None => Future.successful(None)
+          }
+        }
+    }
+  }
+
+  //  // TODO: look up try monad instead of try catch
+  //  // TODO: the e Exception
+  def getImageData(url: String): Future[Option[Array[Byte]]] = Future {
+    try {
```
annebyrne: Could this be returned as an `Either[Success, Error]` instead of using `Try` ?

src/test/scala/com/gu/kindlegen/BookSectionPageSuite.scala
```diff
@@ -10,12 +10,11 @@ import DateUtils._
 class BookSectionPageSuite extends FunSuite {
 
   val capiDate = CapiModelEnrichment.RichJodaDateTime(formatter.parseDateTime("20170724")).toCapiDateTime
-  val ta = Article(TestContent("", "", 1, "", "", capiDate, capiDate, capiDate, "", "", "").toContent)
+  val ta = Article((TestContent("", "", 1, "", "", capiDate, capiDate, capiDate, "", "", "", None, 0).toContent), 0)
```
annebyrne: what is a TA?

src/test/scala/com/gu/kindlegen/ManifestSuite.scala
```diff
@@ -56,11 +23,32 @@ class SectionManifestSuite extends FlatSpec {
         case (l, m, n) => ta.copy(sectionName = l, newspaperBookSection = m, newspaperPageNumber = n)
       }
   }
-  val time = dtFormatter.parseDateTime("20170519011102")
+
+  val time = DateTime.now()
+
+  "SectionManifest.apply" should "convert a sequence of articles to a section Manifest (Contents page)" in {
+
+    assert(SectionManifest(articles, time) === SectionManifest(
```
annebyrne: Personal preference, but the majority of our projects use matchers rather than `assert` as it allows for more expressive checking of results. See here: http://www.scalatest.org/user_guide/using_matchers. For example:

```
SectionManifest(articles, time) should equal SectionManifest(
  value 1,
  value 2,
  .....
  ....
)
```

src/main/scala/com/gu/kindlegen/Article.scala
tomrf1: Looks like a for/yield is needed here

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -30,34 +34,66 @@ object Querier {
...
     response.results
   }
 
-  def responseToArticles(response: Seq[com.gu.contentapi.client.model.v1.Content]): Seq[Article] = {
-    response.map(responseContent =>
-      Article(responseContent))
+  // TODO: handle the possibility of there being no content in the getPrintSentResponse method above
+  def responseToArticles(response: Seq[Content]): Seq[Article] = {
+    val sortedContent = Querier.sortContentByPageAndSection(response)
+    val contentWithIndex = sortedContent.view.zipWithIndex.toList
```
tomrf1: I don't think you need the `view` here. Also, it's nicer to pass in 2 named params than deal with a tuple. How about:
`sortedContent.zipWithIndex.map { case (content, idx) => Article.apply(content, idx) }`

And then:
`def apply(content: Content, index: Int): Article = ...`

src/main/scala/com/gu/kindlegen/Article.scala
regiskuckaertz: the absence of an applicative map operator in scala is lamenting 😩

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -30,34 +34,66 @@ object Querier {
...
+    contentWithIndex.map(Article.apply)
+  }
+
+  def sortContentByPageAndSection(response: Seq[Content]): Seq[Content] = {
+    response.sortBy(content => (content.fields.flatMap(_.newspaperPageNumber), content.tags.find(_.`type` == NewspaperBookSection).get.id))
   }
 
+  // This will probably only be called only when sending the files to Amazon because the images can be stored in memory until they are written to the ftp server. Will use the Article.fileID to name the retrieved image byte[].
+  def fetchImageData(articles: List[Article]): List[Future[Option[Array[Byte]]]] = {
```
regiskuckaertz: Having a single promise to wait for might be more tractable: `fetchImageData: List[Article] => Future[List[Option[Array[Byte]]]]`.
```
Future.sequence { articles map _.imageUrl.map(Querier.getImageData _) }
```

PR #10, opened 2018-01-17
Upgrade dependencies and use config files

hosamaly:
This started as just upgrading dependencies while trying out the SDLC with GitHub, TeamCity, etc. Then I noticed that it depended on the local file system structure, making it impossible to run on TeamCity. I took some time to fix that dependency.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/Main.scala
+++ b/src/main/scala/com/gu/kindlegen/Main.scala
@@ -1,7 +1,23 @@
 package com.gu.kindlegen
 
+import scala.util.{ Failure, Success, Try }
+
 object Main extends App {
-  KindleGenerator.getNitfBundleToDisk
-  println("Done!")
-  // Why does the program not exit here?
+  Settings.load match {
+    case Success(settings) => run(settings)
+    case Failure(error) => handleConfigError(error)
+  }
+
+  private def run(settings: Settings): Unit = {
+    val kindleGenerator = new KindleGenerator(settings)
+    kindleGenerator.getNitfBundleToDisk
+    println("Done!")
+    // Why does the program not exit here?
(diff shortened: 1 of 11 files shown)


src/main/scala/com/gu/kindlegen/Main.scala
```diff
@@ -1,7 +1,23 @@
...
+    case Success(settings) => run(settings)
+    case Failure(error) => handleConfigError(error)
+  }
+
+  private def run(settings: Settings): Unit = {
+    val kindleGenerator = new KindleGenerator(settings)
+    kindleGenerator.getNitfBundleToDisk
+    println("Done!")
+    // Why does the program not exit here?
+  }
```
hosamaly: Nothing changed in this method; it's the same as the body of this class used to be.

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -77,12 +64,12 @@ object Querier {
   }
 
   def sortContentByPageAndSection(response: Seq[Content]): Seq[Content] = {
-    response.sortBy(content => (content.fields.flatMap(_.newspaperPageNumber), content.tags.find(_.`type` == NewspaperBookSection).get.id))
+    response.sortBy(content => (content.fields.flatMap(_.newspaperPageNumber), content.tags.find(_.`type` == NewspaperBookSection).map(_.id)))
```
hosamaly: This is a tiny side-fix to avoid issues when there is no content with the expected tag.

mchv: 👍

mchv:
Good work @hosamaly

PR #11, opened 2018-02-19
Upgrade CAPI client

hosamaly:
And migrate from Joda-Time to Java 8 time


src/main/scala/com/gu/kindlegen/DateUtils.scala
```diff
@@ -1,31 +1,80 @@
...
+  private def editionDateTime: DateTime = formatterWithDashes.parseDate("2017-05-19") // to have a date I know the results for
   //  def editionDateTime: DateTime = DateTime.now()
 
-  def editionDateString: String = formatterWithDashes.print(editionDateTime)
-  def editionDateStart: DateTime = DateTime.parse(editionDateString).withMillisOfDay(0).withMillisOfSecond(0)
-  def editionDateEnd: DateTime = DateTime.parse(editionDateString).withHourOfDay(23).withMinuteOfHour(59).withSecondOfMinute(59).withMillisOfSecond(999)
+  def editionDateStart: DateTime = editionDateTime.startOfDay
+  def editionDateEnd: DateTime = editionDateTime.endOfDay
+
+  /* Adaptor classes to simplify the migration from Joda-Time to Java 8 Time */
```
tomrf1: As discussed this is a temporary measure to keep the project compiling before a rewrite

hosamaly: Temporary classes are no longer there.

PR #13, opened 2018-04-10
Move XML validation to the main source

hosamaly:
We'd like to validate the NITF that we intend to generate from CAPI. This PR encapsulates and cleans up the logic to validate XML against a schema.

Additionally, this PR includes moving the logic to transform XHTML into NITF to the main source as mentioned in the previous PR.


LATaylor-guardian:
Looks good 👍

PR #14, opened 2018-04-17
Fix search

hosamaly:
This PR fixes the searching logic to do a proper query against the print-sent content.

It also adds a few test cases to cover existing logic.


src/main/resources/reference.conf
```diff
@@ -1,4 +1,4 @@
 content-api {
   key = "Get a key from https://bonobo.capi.gutools.co.uk"
-  url = "https://content.guardianapis.com/"
+  url = "https://preview.content.guardianapis.com"
```
LATaylor-guardian: This is a public repository and as such we do not want to include this URL in the repo. This should be passed in as an application secret - please remove and clean up the history.

src/main/scala/com/gu/contentapi/client/GuardianContentClientFactory.scala
```diff
@@ -0,0 +1,28 @@
...
+  def apply(apiKey: String,
+            apiServiceUrl: URI = DefaultServiceUrl): GuardianContentClient = {
+    require(apiKey.nonEmpty, "API key is empty")
+    require(apiServiceUrl.isAbsolute, "URI is not absolute")
+    require(Option(apiServiceUrl.getRawQuery).forall(_.isEmpty), "URI specifies query parameters")
+
+    build(apiKey, apiServiceUrl.toASCIIString)
+  }
+
+  def internal: GuardianContentClient = build("", "https://internal.content.guardianapis.com")
```
LATaylor-guardian: Same as above - this URL is considered a secret.

src/main/scala/com/gu/contentapi/client/GuardianContentClientFactory.scala
```diff
@@ -0,0 +1,28 @@
...
+  def preview: GuardianContentClient = build("", "https://preview.content.guardianapis.com")
```
LATaylor-guardian: Here too.

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -46,37 +46,43 @@ object Article {
...
-    val index = contentWithIndex._2
-    //    for {
-    //      newspaperBookSection <- content.tags.find(_.`type` == NewspaperBookSection)
-    //    } Article(
+  def apply(content: Content, index: Int): Article = {
+    val maybeSectionTag = content.tags.find(_.`type` == NewspaperBookSection)
+    val maybeNewspaperDate = content.fields.flatMap(_.newspaperEditionDate).orElse(content.webPublicationDate)
+
+    val contentId = s"Content(${content.id})"
+    require(content.fields.nonEmpty, s"$contentId retrieved without fields")
```
LATaylor-guardian: This feels like a non-Scala way of handling whether something is there or not. Can we have the apply method return an Option[Article] so that we don't need these requires, causing an exception to be thrown. This will also mean the `.get`'s can be removed below.

hosamaly: This is an interesting discussion. Although I wanted to minimise the changes here, for now, I agree that this isn't the best implementation.

The reason these exist as assertions is that they are part of the contract of the print-sent query.

I don't think returning an `Option[Article]` is the best solution, though, because it doesn't tell the caller anything about the issues that prevented the creation of the object. I think that the use of Cats' `Validated` (or a similar pattern) is a better option. However, I don't want to go down this route until we clean up a few other things (e.g. logging).

If you think this is urgent, then I'm happy to change the method to return an `Option[Article]` until we get back to this.

LATaylor-guardian: No simply an observation - feel free to continue.

src/main/scala/com/gu/contentapi/client/model/PrintSentSearchQuery.scala
```diff
@@ -0,0 +1,21 @@
...
+    new PrintSentSearchQuery(parameterHolder).tier("internal")
+  }
+}
+
+final case class PrintSentSearchQuery private(parameterHolder: Map[String, Parameter])
+  extends SearchQueryBase[PrintSentSearchQuery]
+     with TierParameter[PrintSentSearchQuery] {
+
+  override def pathSegment: String = "content/print-sent"
+  override def withParameters(parameterMap: Map[String, Parameter]) = new PrintSentSearchQuery(parameterMap)
```
regiskuckaertz: Shouldn't this be `= PrintSentSearchQuery(parameterMap)` instead? so that the `tier` is correctly set.

hosamaly: Great catch! 👏 🦅

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
```diff
@@ -1,12 +1,14 @@
...
+import java.nio.file.{Files, Path}
+import java.time.LocalDate
 
 import scala.concurrent.Await
 import scala.concurrent.duration._
 
-class KindleGenerator(settings: Settings) {
-  val Querier = new Querier(settings)
+class KindleGenerator(settings: Settings, editionStart: LocalDate) {
+  import scala.concurrent.ExecutionContext.Implicits.global
```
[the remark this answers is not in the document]
hosamaly: Maybe. In this design (with which I'm unhappy!), this is the main entry point. I'll consider changing this when I get to redesigning this class.

PR #16, opened 2018-04-19
Improve design

hosamaly:
Kindly check the commits for details.


src/main/scala/com/gu/kindlegen/KindleGenerator.scala
```diff
@@ -11,22 +11,12 @@ class KindleGenerator(settings: Settings, editionStart: LocalDate) {
...
-      Querier.getAllArticleImages(articles),
-      120.seconds
-    )
+    val articles = Querier.fetchAllArticles()
 
-    val files = articles.map(articleToFile) ++ images.map(articleImageToFile)
-    files
+    articles.zipWithIndex.flatMap { case (article, index) =>
+      Some(articleToFile(article, index)) ++
+        Querier.getArticleImage(article).map(Await.result(_, 15.seconds)).map(articleImageToFile(_, index))
```
regiskuckaertz: If we don't need to acknowledge timeouts, then it might be more idiomatic to assimilate it as `None`.

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -25,60 +30,49 @@ object Querier {
...
     val futures = articles.flatMap(getArticleImage)
     Future.sequence(futures)
   }
 
-  def getArticleImage(article: Article): Option[Future[ArticleImage]] = {
-    article.imageUrl.map(url => {
-      val future = getImageData(url)
-      future.map(bytes => ArticleImage(articleId = article.fileId, data = bytes))
-    })
+  def getArticleImage(article: Article): Option[Future[ImageData]] = {
```
regiskuckaertz: In my opinion, that should be `Future[Option[ImageData]]` instead.

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -25,60 +30,49 @@ object Querier {
...
       case Success(article) => article
+      // TODO log the issue in the case of failure
     }
-  }
+  )
 
   // This will probably only be called only when sending the files to Amazon because the images can be stored in memory until they are written to the ftp server. Will use the Article.fileID to name the retrieved image byte[].
 
-  def getAllArticleImages(articles: Seq[Article]): Future[Seq[ArticleImage]] = {
+  def getAllArticleImages(articles: Seq[Article]): Future[Seq[ImageData]] = {
```
regiskuckaertz: Because `sequence . map f == traverse f`, you can skip the intermediate step:
```scala
Future.traverse(articles)(getArticleImage(_)).map(_.flatten)
```

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -73,8 +74,9 @@ class Querier(settings: Settings, editionDate: LocalDate)(implicit ec: Execution
 
   def getArticleImage(article: Article): Option[Future[ArticleImage]] = {
     article.imageUrl.map(url => {
+      val extension = url.substring(url.lastIndexOf('.') + 1)
```
regiskuckaertz: I've seen this somewhere else in this PR, may be worth putting somewhere unique

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
regiskuckaertz: Scratch that, I saw your commit message where you say all `Await`s will go away

src/main/scala/com/gu/kindlegen/KindlePublishingSearchQuery.scala
```diff
@@ -6,21 +6,21 @@ import java.time.temporal.ChronoUnit.DAYS
 import com.gu.contentapi.client.model.PrintSentSearchQuery
 
 object KindlePublishingSearchQuery {
-  // pagination is not necessary for a print-sent query; it only makes it slower and there are only ~200 per day
-  val AllResultsPageSize = 500
+  // pagination is not necessary nor desired for a print-sent query
+  // it only makes it slower and there are only ~200 per day
+  // additionally, `newspaper-edition` date is always set to midnight, resulting in duplicate results across pages
+  val AllResultsPageSize = 400  // maximum page size accepted by CAPI
```
tomrf1: Nice, I didn't realise PrintSent allows 400!

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -1,86 +1,61 @@
...
-
+  def getBodyBlocks(content: Content): Seq[String] = {
     val blocks = content.blocks
-    val bodyBlocks: Seq[Block] = blocks.flatMap(_.body).getOrElse(Nil)
-    val bodyBlocksElements: Seq[BlockElement] = bodyBlocks.flatMap(_.elements)
-    val textElemField: Seq[String] = bodyBlocksElements.flatMap(_.textTypeData.flatMap(_.html))
-    val noOpt = textElemField.mkString
-    noOpt
+    val bodyBlocks = blocks.flatMap(_.body).getOrElse(Nil)
+    // TODO what if bodyBlocks.length < blocks.totalBodyBlocks ?
```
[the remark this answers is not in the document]
hosamaly: That's good to know. Thanks.

src/main/scala/com/gu/kindlegen/KindlePublishingSearchQuery.scala
hosamaly: I couldn't find out where that 400 is specified. I found it by trial and error 😊

src/main/scala/com/gu/kindlegen/Querier.scala
hosamaly: Agreed 👍

src/main/scala/com/gu/kindlegen/KindlePublishingSearchQuery.scala
tomrf1: in case you're interested :) https://github.com/guardian/content-api/blob/a943560bc63d0bbce43c74d2fe69689b7ebdc8a2/concierge/src/main/scala/com.gu.contentapi.concierge/web/ConciergeService.scala#L107

src/main/scala/com/gu/kindlegen/Querier.scala
hosamaly: Done

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
hosamaly: This should be fixed in e9fed92.

src/main/scala/com/gu/kindlegen/Querier.scala
hosamaly: I have been thinking about this. I am now leaning more towards `Option[Future[ImageData]]`. That's because the existence of the value is known before creating the future. Semantically (and performance-wise), `Option[Future[_]]` more accurately describes the situation.

hosamaly: On the other hand, it's easier to use a `Future[Option[_]]` than an `Option[Future[_]]`!

regiskuckaertz: Yes, I understand your reasoning, but it seems to me that is an artefact of the data structure `Article`, not of the semantics of `getArticleImage`. If you abstracted, i.e. `Querier[Article]`, then you wouldn't be able to rely on that assumption and so `Future[Option[_]]` would be the only option. Agreed, if performance is that different, there is no other choice.

hosamaly: Thanks. I decided to leave it as-is in the end.

tomrf1:
👍 this looks good to me

PR #17, opened 2018-04-26
Generate RSS manifests

hosamaly:
In this PR, the main highlight is that we're now generating an RSS file for each section, listing the articles in that section, and another RSS file that lists the section files.

Additionally, a number of properties have been moved to the configuration file.


src/main/scala/com/gu/kindlegen/Link.scala
```diff
@@ -0,0 +1,73 @@
...
+      new RelativePath(source.toString, relativeTo)
+
+    def apply(source: String, relativeTo: AbsolutePath): Try[RelativePath] =
+      Try(Paths.get(source)).map(_ => new RelativePath(source, relativeTo))
+
+    def from(source: Path, relativeTo: AbsolutePath): RelativePath = apply(source, relativeTo)
+    def from(source: String, relativeTo: AbsolutePath): RelativePath = apply(source, relativeTo).get
+  }
+
+  abstract class Factory[T <: Link, Mirror](constructor: String => T, mirrorConstructor: String => Mirror) {
```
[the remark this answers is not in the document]
hosamaly: You could say so. It's the type that is used for validation, to be honest. I struggled to come up with a good name.

src/main/scala/com/gu/kindlegen/BookSection.scala
```diff
@@ -1,17 +1,40 @@
...
+  /** Groups articles into book sections, sorted according to each section's articles' page number */
+  def fromArticles(articles: Seq[Article]): Seq[BookSection] = {
+    articles.groupBy(_.section)
+      .values.map(apply)(breakOut)
+      .sortBy(section => (section.firstPageNumber, section.lastPageNumber, section.title))
+  }
+
+  /** Creates a book section from a collection of articles that all belong to the same section */
+  def apply(articles: Seq[Article]): BookSection = {
+    require(articles.nonEmpty, "A book section must have at least one article!")
```
tomrf1: I'm not entirely opposed to the use of `require`, I can see why it's convenient, particularly in a lambda. It does worry me a bit to see functions that can throw exceptions rather than returning e.g. Try or Either, as it relies on the caller to know this.
Don't let this delay getting the program working, but it's something to consider later on perhaps

src/main/scala/com/gu/kindlegen/Querier.scala
```diff
@@ -41,16 +41,17 @@ class Querier(settings: Settings, editionDate: LocalDate)(implicit ec: Execution
...
+  def sortedArticles(results: Seq[Content]): Seq[Article] = {
+    val articles = results.map { content => Try(Article(content)) }.collect {
       case Success(article) => article
       // TODO log the issue in the case of failure
     }
-  )
+    sortArticlesByPageAndSection(articles)
+  }
+
+  private def sortArticlesByPageAndSection(articles: Seq[Article]): Seq[Article] = {
```
regiskuckaertz: What about defining an ordering:
```scala
implicit private val articleOrdering = Ordering.by[Article, (Int, Int)](a => (a.newspaperPageNumber, a.sectionId))
```
then you can use whatever sort algorithm you need, e.g. `Sorting.stableSort(articles)`

src/main/scala/com/gu/kindlegen/BookSection.scala
hosamaly: Thanks. I don't think `Try` and `Either` are very useful, unless we resort to using suppressed exceptions or `Either[T, Seq[Throwable]]`. I honestly think that Scala should add something like Cats' `Validated` to the standard library. Otherwise, we'll keep having to [reinvent the wheel](https://github.com/guardian/kindle-gen/blob/a16decdeffd40694da3e6a4a82543a58298b624f/src/main/scala/com/gu/xml/validation/XmlValidationResult.scala).

On the other hand, I agree with you. Relying on the caller to know that exceptions may be thrown is an undesired pattern. I agree that I should consider it later.

src/main/scala/com/gu/kindlegen/Querier.scala
hosamaly: I hadn't known about [`Sorting`](https://www.scala-lang.org/files/archive/api/2.12.6/scala/util/Sorting$.html). Thanks!

This method no longer exists but I'll keep this in mind in the future.

src/main/scala/com/gu/kindlegen/Section.scala
```diff
@@ -0,0 +1,13 @@
...
+import com.gu.contentapi.client.model.v1.Tag
+import com.gu.kindlegen.Link.AbsoluteURL
+
+
+case class Section(id: String, title: String, link: Link)
+
+object Section {
+  def apply(tag: Tag): Section = Section(tag.id, tag.webTitle, AbsoluteURL.from(tag.webUrl))
+
+  implicit val ordering: Ordering[Section] = Ordering.by(section => (section.id, section.title))
```
regiskuckaertz: Can two sections share the same identifier? If yes, you could similarly define `equals` to only compare ids.

src/test/scala/com/gu/kindlegen/SectionsManifestSpec.scala
```diff
@@ -20,8 +20,8 @@ class SectionsManifestSpec extends FlatSpec {
       ("Top Stories", "theguardian/mainsection/topstories", 4),
       ("Top Stories", "theguardian/mainsection/topstories", 4)
     ).map {
-
-        case (l, m, n) => ta.copy(sectionName = l, sectionId = m, newspaperPageNumber = n)
+        case (title, id, page) => ta.copy(newspaperPageNumber = page,
+          section = Section(id = id, title = title, link = null))
```
regiskuckaertz: Does `null` need to be lifted into your `Link` ADT to represent the null anchor (`#`)? Feel free to ignore, it looks like it's only useful for your tests.

src/test/scala/com/gu/kindlegen/ArticleNITFSuite.scala
```diff
@@ -15,6 +15,7 @@ class ArticleNITFSuite extends FunSuite {
       newspaperPageNumber = 2,
       title = "my title",
       docId = "section/date/title",
+      link = Link.DataURI.from("data:,"),
```
regiskuckaertz: Another instance that looks similar to `null` on principle in the `Section` tests.

src/main/scala/com/gu/kindlegen/Section.scala
hosamaly: I assume that section identifiers should be unique but there is no way of guaranteeing this here.

tomrf1: yes, the sectionId is unique

src/test/scala/com/gu/kindlegen/SectionsManifestSpec.scala
hosamaly: This was an intermediate state. There are no more `null` links now.

hosamaly: For the purposes of this codebase, the null anchor is disadvantageous. All `Link` classes, in the current design, represent an absolute link that can be used on its own. Perhaps I should write this in a doc comment.

src/test/scala/com/gu/kindlegen/ArticleNITFSuite.scala
hosamaly: This was an intermediate state. There are no more links like this now.

src/main/scala/com/gu/kindlegen/Section.scala
hosamaly: It's unique in the Content API. However, there is no guarantee that this codebase creates a single section object for each section.

To be honest, I don't think it matters so I'm happy to remove the title from the ordering if you think I should.

regiskuckaertz: > there is no guarantee that this codebase creates a single section object

That is a concern worth addressing in the factory for sections then, not in the ordering. As you said, that is trivial.

src/main/scala/com/gu/kindlegen/BookSection.scala
regiskuckaertz: I guess each covers different, albeit overlapping, semantic domains. Exceptions (`Try`) should only ever be used for things that are not under the application's control. `Either` covers so many other uses cases (as a bifunctor, for instance), the right-biased aspect is just convention coming from Haskell.

hosamaly: I think that the right-biased nature is because it's easier to remember that _"`Right` is right"_ :)

tomrf1: If you're just looking to log error strings then `Either[String,T]` is fine here, no?

hosamaly: It probably is. Let's revisit it when we've got a bit more time.

PR #18, opened 2018-05-01
Configurable Settings

hosamaly:
Make settings configurable; avoid hard-coded settings.


src/main/resources/reference.conf
```diff
@@ -3,11 +3,24 @@ content-api {
   url = "https://content.guardianapis.com"
 }
 
+query {
+  sectionTagType = "NewspaperBookSection"  // equal to TagType#name
```
hosamaly: I wrote this in the commit:
> Using `NewspaperBook` would produce less sections while using `NewspaperBookSection` would produce more. We can also create custom publications using `Keword` or `Contributor`.
Do you think I should include the comment either in the code or in the configuration file?

src/main/resources/reference.conf
```diff
@@ -3,11 +3,24 @@ content-api {
...
+  downloadTimeout = 10 minutes  // it takes a long time to download master images
```
tomrf1: can/should a lambda run for this long?

src/main/resources/reference.conf
tomrf1: Which tag type does the old app use?

src/main/resources/reference.conf
hosamaly: No, it can't run for more than 5 minutes, and it shouldn't. I put this here temporarily because TeamCity test builds were timing out. I should move it to `test/reference.conf`.

src/main/resources/reference.conf
hosamaly: `NewspaperBookSection`

src/main/resources/reference.conf
tomrf1: 👍

hosamaly: Moved to test config in 661db7b.

PR #19, opened 2018-05-01
Generate RSS manifests

hosamaly:
Clean up and test the generation of RSS manifests, and make sure XML is escaped properly.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/Article.scala
+++ b/src/main/scala/com/gu/kindlegen/Article.scala
@@ -19,8 +19,6 @@ object Article {
   def getBodyBlocks(content: Content): Seq[String] = {
     val blocks = content.blocks
     val bodyBlocks = blocks.flatMap(_.body).getOrElse(Nil)
-      .ensuring(bodies => blocks.flatMap(_.totalBodyBlocks).forall(_ == bodies.length))
-
     val bodyElements = bodyBlocks.flatMap(_.elements).filter(_.`type` == ElementType.Text)
     val htmlBlocks = bodyElements.flatMap(_.textTypeData.flatMap(_.html))
     htmlBlocks
--- a/src/main/scala/com/gu/kindlegen/Article.scala
+++ b/src/main/scala/com/gu/kindlegen/Article.scala
@@ -19,7 +19,8 @@ object Article {
   def getBodyBlocks(content: Content): Seq[String] = {
     val blocks = content.blocks
     val bodyBlocks = blocks.flatMap(_.body).getOrElse(Nil)
-    // TODO what if bodyBlocks.length < blocks.totalBodyBlocks ?
+      .ensuring(bodies => blocks.flatMap(_.totalBodyBlocks).forall(_ == bodies.length))
+
     val bodyElements = bodyBlocks.flatMap(_.elements).filter(_.`type` == ElementType.Text)
     val htmlBlocks = bodyElements.flatMap(_.textTypeData.flatMap(_.html))
     htmlBlocks
@@ -40,7 +41,7 @@ object Article {
   def apply(content: Content, newspaperDate: CapiDateTime, fields: ContentFields, sectionTag: Tag): Article = {
     Article(
       section = Section(sectionTag),
-      newspaperPageNumber = fields.newspaperPageNumber.getOrElse(0),
+      newspaperPageNumber = fields.newspaperPageNumber.getOrElse(Int.MaxValue),  // move to the end of the section
       title = content.webTitle,
       docId = content.id,
       link = Link.AbsoluteURL.from(content.webUrl),
--- a/src/main/scala/com/gu/kindlegen/BookSection.scala
+++ b/src/main/scala/com/gu/kindlegen/BookSection.scala
@@ -1,6 +1,10 @@
 package com.gu.kindlegen
 
+import java.time.LocalDate
+
 import scala.collection.breakOut
+
+import com.gu.contentapi.client.utils.CapiModelEnrichment._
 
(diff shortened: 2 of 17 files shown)


src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -19,7 +19,8 @@ object Article {
   def getBodyBlocks(content: Content): Seq[String] = {
     val blocks = content.blocks
     val bodyBlocks = blocks.flatMap(_.body).getOrElse(Nil)
-    // TODO what if bodyBlocks.length < blocks.totalBodyBlocks ?
+      .ensuring(bodies => blocks.flatMap(_.totalBodyBlocks).forall(_ == bodies.length))
```
tomrf1: I don't think this check is necessary - capi should always return all blocks if requested

src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -40,7 +41,7 @@ object Article {
   def apply(content: Content, newspaperDate: CapiDateTime, fields: ContentFields, sectionTag: Tag): Article = {
     Article(
       section = Section(sectionTag),
-      newspaperPageNumber = fields.newspaperPageNumber.getOrElse(0),
+      newspaperPageNumber = fields.newspaperPageNumber.getOrElse(Int.MaxValue),  // move to the end of the section
```
tomrf1: how is this field used? I'm guessing it's not included in the output

src/main/scala/com/gu/kindlegen/Rss.scala
```diff
@@ -0,0 +1,44 @@
...
+      </channel>
+    </rss>
+}
+
+object RssManifest {
+  private final val Encoding = "UTF-8"
+
+  private val formatter = DateTimeFormatter
+    .ofPattern("EEE, dd MMM yyyy HH:mm:ss Z")  // Tue, 01 May 2018 17:10:54 +0000
+    .withZone(ZoneId.of("UTC"))
```
regiskuckaertz: There's a constant if you prefer: `ZoneOffset.UTC`

src/main/scala/com/gu/kindlegen/Article.scala
hosamaly: I guess this is a question of how defensive / trusting this code should be. This assertion may not guard against CAPI as much as it defends against an incorrect query where `show-blocks` is missing or set to an undesired value (e.g. `show-blocks=main`).

Do you think this is the wrong place to have this check?

src/main/scala/com/gu/kindlegen/Article.scala
hosamaly: It's used in `BookSection` to [sort sections](https://github.com/guardian/kindle-gen/blob/a9d147154bc9b1b116568df271e4b7c464ae78b5/src/main/scala/com/gu/kindlegen/BookSection.scala#L30) and to [sort articles in a section](https://github.com/guardian/kindle-gen/blob/a9d147154bc9b1b116568df271e4b7c464ae78b5/src/main/scala/com/gu/kindlegen/BookSection.scala#L43).

src/main/scala/com/gu/kindlegen/Rss.scala
hosamaly: Thanks. I hadn't realised that `ZoneOffset extends ZoneId`. I'll change it.

src/main/scala/com/gu/kindlegen/BookSection.scala
```diff
@@ -13,8 +17,9 @@ case class BookSection(section: Section, articles: Seq[Article]) extends Linkabl
   def id: String = section.id
   def title: String = section.title
   def link: Link = section.link
+  lazy val publicationDate: LocalDate = articles.map(_.pubDate).minBy(_.dateTime).toOffsetDateTime.toLocalDate
 
-  def publicationDate = articles.map(_.pubDate).minBy(_.dateTime)
+  def withLink(newLink: Link): BookSection = copy(section = section.copy(link = newLink))
```
regiskuckaertz: It seems you are using this solely for the purpose of collecting local file names at the end of your compilation. You end up doing copies of your value objects all the way down ([here](https://github.com/guardian/kindle-gen/blob/a9d147154bc9b1b116568df271e4b7c464ae78b5/src/main/scala/com/gu/kindlegen/KindleGenerator.scala#L64), [here](https://github.com/guardian/kindle-gen/blob/a9d147154bc9b1b116568df271e4b7c464ae78b5/src/main/scala/com/gu/kindlegen/KindleGenerator.scala#L75) and [here](https://github.com/guardian/kindle-gen/blob/a9d147154bc9b1b116568df271e4b7c464ae78b5/src/main/scala/com/gu/kindlegen/KindleGenerator.scala#L97)). That may create a lot of GC pressure as objects are created and consumed immediately -- I don't know how good Scala is at structural sharing. Getting around that is not very hard, so if you feel like it I can give you the gist of it.

src/main/scala/com/gu/kindlegen/Article.scala
tomrf1: the `totalBodyBlocks` field is only added if the client has requested body blocks, so this test is unnecessary - https://github.com/guardian/content-api/blob/c93ebc8e55b9270542a66d9a66ce33521e6984b5/concierge/src/main/scala/com.gu.contentapi.concierge/processors/ContentProcessor.scala#L179

hosamaly: I removed the check in 0a50ec3.

PR #12, opened 2018-04-04
HTML Transformation to NITF

hosamaly:
The main class in this PR is `NitfValidator`, the companion object of which has logic to transform existing NITF files (which typically contain the HTML `body` of articles) to become valid NITF files.

The transformations should make it to the main source code shortly to be used after cleaning the HTML using Jsoup. They should help guarantee that we produce valid NITF (according to Amazon's schema).

This test was run against ~27k files generated between 18/08/2017 and 03/04/2018.

PR #20, opened 2018-05-03
Produce Valid NITF

hosamaly:
The purpose of this PR is to make sure that the NITF that we generate is valid.

We got rid of using strings and instead use scala-xml to make sure strings are properly escaped. We also use Jsoup to make sure that the HTML is valid. Finally, `XhtmlToNitfTransformer` (introduced in PRs #12 and #13) maps the XHTML to a valid NITF structure.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/ArticleNITF.scala
+++ b/src/main/scala/com/gu/kindlegen/ArticleNITF.scala
@@ -10,6 +10,7 @@ import org.jsoup.nodes.Document.OutputSettings.Syntax
 import org.jsoup.nodes.Entities.EscapeMode
 
 import com.gu.kpp.nitf.XhtmlToNitfTransformer
+import com.gu.xml._
 
 
 object ArticleNITF {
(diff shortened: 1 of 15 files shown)


src/main/scala/com/gu/kindlegen/ArticleNITF.scala
```diff
@@ -1,35 +1,97 @@
...
-     |<byline>${article.byline}</byline>
-     |<abstract>${article.articleAbstract}</abstract>
-     |</body.head>
-     |<body.content>${article.bodyBlocks.mkString("\n") /* TODO convert to NITF blocks*/}</body.content>
-     |<body.end />
-     |</body>
-     |</nitf>""".stripMargin
+  private val cleaner = new Cleaner(Whitelist.relaxed  // TODO remove unwanted tags
+    .addProtocols("a", "href", "#")
+    .removeProtocols("a", "href", "ftp", "http", "https", "mailto")
```
tomrf1: is this undoing the work of line 23?

src/main/scala/com/gu/kpp/nitf/XhtmlToNitfTransformer.scala
```diff
@@ -5,21 +5,21 @@ import scala.xml._
...
-    * @param xhtmlBody an element representing the <body> of an XHTML document
-    * @return an element representing the <body.content> of an equivalent NITF document
+    * @param xhtml an element representing an XHTML document
+    * @return the element transformed to match NITF specs
     */
-  def apply(xhtmlBody: Elem): Elem = {
-    val xhtml = if (xhtmlBody.label == "body") xhtmlBody.copy(label = "body.content") else xhtmlBody
-    xhtml.transform(transformationRules).toElem()
+  def apply(xhtml: Elem): Elem = {
+    xhtml.transform(transformationRules).asInstanceOf[Elem]//.toElem()
```
tomrf1: why has this line changed?

hosamaly: I replaced `toElem()` with `asInstanceOf` while debugging an issue. I should revert it. (But the first line is correctly removed.)

hosamaly:
Thanks!

PR #21, opened 2018-05-04
Lambda infrastructure

hosamaly:
Configure the app to run as an AWS Lambda


src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,63 @@
...
-    Option(System.getenv("Stage")).getOrElse("DEV")
-  )
-}
+import awscala._
+import awscala.s3._
+import com.amazonaws.services.lambda.runtime.Context
 
 object Lambda {
+  // env.AWS_REGION -> eu-west-1
+  // env.Stage -> CODE
```
hosamaly: I'm still going to work on cleaning up this file.

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,63 @@
...
+        val fileSettings = settings.publishing.files
+        val originalOutputDir = fileSettings.outputDir.toAbsolutePath
+        val customFileSettings = fileSettings.copy(originalOutputDir.resolve(date.toString))
+        val customSettings = settings.withPublishingFiles(customFileSettings)
+        val kindleGenerator = KindleGenerator(customSettings, date)
+
+        logger.log(s"Generating for $date; writing to ${customFileSettings.outputDir}")
+        val files = kindleGenerator.writeNitfBundleToDisk()
+
+        files.par.foreach { path =>
```
hosamaly: I'm thinking of collecting all files into an archive and uploading that single archive. If we do this then we'll have to uncompress them again when we SFTP them.

cfn.yaml
```diff
@@ -49,6 +57,14 @@ Resources:
               Action:
                 - lambda:InvokeFunction
               Resource: "*"
+        - PolicyName: publishToS3
+          PolicyDocument:
+            Statement:
+              Effect: Allow
+              Action:
+                - s3:PutObject
+                - s3:ListAllMyBuckets  # AWScala uses it to find the bucket
```
hosamaly: This is not the right syntax; I need to fix this!

tomrf1: It is odd that this library requires `ListAllMyBuckets` permission, since you already know the name of the bucket. Can this be avoided?

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,63 @@
...
+      params.get("date").map(_.toString).map(LocalDate.parse)
+        .orElse(params.get("time").map(_.toString).map(Instant.parse(_).atZone(UTC).toLocalDate))  // scheduled event
+        .getOrElse(LocalDate.now)
+
+    val stage = sys.env("Stage")  // CODE or PROD
+    val region = Region(sys.env("AWS_REGION"))
+    val bucketName = "kindle-gen-published-files"
+
+    implicit val s3: S3 = S3.at(region)
+    val bucket = s3.bucket(bucketName).getOrElse(throw new RuntimeException("Cannot find bucket"))
```
tomrf1: I'm not sure how much we gain from using this library for s3.
I think you can achieve the same with the following, unless there was a feature you particularly wanted?
```
val s3Client: AmazonS3 = AmazonS3ClientBuilder.standard().withRegion(region).build()
...
s3Client.putObject(bucket, path, file)
```

src/main/scala/com/gu/kindlegen/Lambda.scala
tomrf1: sounds sensible

riff-raff.yaml
```diff
@@ -1,19 +1,23 @@
...
+regions:
+  - eu-west-1
 
 deployments:
   kindle-gen:
     type: aws-lambda
     parameters:
-      bucket: CHANGE_ME-dist
+      bucket: content-api-dist
       functionNames: [kindle-gen-]
```
LATaylor-guardian: Is this correct?

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,63 @@
...
-    logger.log(s"Starting $env")
-    logger.log(process(lambdaInput.name, env))
+
+    val params = parameters.asScala
+    val date =
+      params.get("date").map(_.toString).map(LocalDate.parse)
+        .orElse(params.get("time").map(_.toString).map(Instant.parse(_).atZone(UTC).toLocalDate))  // scheduled event
+        .getOrElse(LocalDate.now)
+
+    val stage = sys.env("Stage")  // CODE or PROD
```
LATaylor-guardian: If we do not have particular config we need that we expect to receive from environment variables then we should fail loudly. E.g

```
Option(sys.env("AWS_REGION")) getOrElse sys.error("")
```

riff-raff.yaml
tomrf1: Yes, we discovered this setting on friday:
`One or more function names to update with the code from fileNameParam. Each function name will be suffixed with the stage, e.g. MyFunction- becomes MyFunction-CODE`

src/main/scala/com/gu/kindlegen/Lambda.scala
tomrf1: I don't think so, my understanding is the call here to `.par.foreach` blocks until completion

cfn.yaml
LATaylor-guardian: Perfect.

src/main/scala/com/gu/kindlegen/Lambda.scala
LATaylor-guardian: Perfect.

cfn.yaml
hosamaly: I would have to submit a PR to fix [the code](https://github.com/seratch/AWScala/blob/0.6.3/src/main/scala/awscala/s3/S3.scala#L63). I'll see if it's easy to do.

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: True, we're only using a tiny subset of the API here. It was just easier to use and the docs were more accessible!

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: Will do.

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: Removed the dependency in 20b5579

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: I've removed the quoted line and made sure we have appropriate defaults for other properties.

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,63 @@
...
+        val originalOutputDir = fileSettings.outputDir.toAbsolutePath
+        val customFileSettings = fileSettings.copy(originalOutputDir.resolve(date.toString))
+        val customSettings = settings.withPublishingFiles(customFileSettings)
+        val kindleGenerator = KindleGenerator(customSettings, date)
+
+        logger.log(s"Generating for $date; writing to ${customFileSettings.outputDir}")
+        val files = kindleGenerator.writeNitfBundleToDisk()
+
+        files.par.foreach { path =>
+          // TODO what if the file already exists in S3?
```
tomrf1: when would this happen? Do the filenames or paths contain the timestamp?

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,63 @@
...
+          val s3path = stage + "/" + originalOutputDir.relativize(path.toAbsolutePath)
+          logger.log(s"Uploading $s3path")
+          s3.putObject(bucketName, s3path, path.toFile)
+          // TODO what if uploading to S3 fails?
```
tomrf1: `putObject` will throw an exception, which will cause the lambda to end. Best to wrap each call in a try and log the error

cfn.yaml
```diff
@@ -27,78 +28,40 @@ Parameters:
...
-          Arn: !GetAtt Lambda.Arn
-
-  DailyEventLambdaPermission:
-    Type: AWS::Lambda::Permission
-    Properties:
-      Action: lambda:InvokeFunction
-      FunctionName: !GetAtt Lambda.Arn
-      Principal: events.amazonaws.com
-      SourceArn: !GetAtt DailyEvent.Arn
+      Events:
```
tomrf1: this is really nice, very concise

cfn.yaml
```diff
@@ -16,72 +18,50 @@ Parameters:
       - CODE
       - PROD
     Default: CODE
+  ContentApiKey:
+    Description: 'A key generated by https://bonobo.capi.gutools.co.uk'
+    Type: String
```
mchv: Please add `NoEcho` so that they key remains a secret, see [documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-specific-parameter-types)

hosamaly: Thanks. Done in ded8b4a.

hosamaly: So, I did this but we still see the value in the Environment Variables section when we run the function. Is there a way around this?

regiskuckaertz: Variables can be encrypted at rest and in transit automatically by using KMS. I think you just need to provide the key's ARN: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-kmskeyarn

cfn.yaml
```diff
@@ -16,72 +18,65 @@ Parameters:
...
+                },
+                ${CustomConfig}
+              }
+      Events:
+        DailyEvent:
+          Type: Schedule
+          Properties:
+            Schedule: cron(30 1 * * ? *)  # Run at 01:30 AM every day
+      CodeUri:
+        Bucket: !Ref DeployBucket
```
regiskuckaertz: I think it's safe to inline `content-api-dist`

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,67 @@
...
+      params.get("date").map(LocalDate.parse)
+        .orElse(params.get("time").map(Instant.parse(_).atZone(UTC).toLocalDate))  // scheduled event
+        .getOrElse(LocalDate.now)
+
+    val config = resolveConfig(sys.env.getOrElse("ConfigSettings", ""))
+    val bucketName = config.getString("serialization.s3.bucket")
+    val pathPrefix = config.getString("serialization.s3.prefix")
+
+    val s3 = AmazonS3ClientBuilder.defaultClient()
+    // TODO should we validate the connection and permissions?
```
regiskuckaertz: My opinion: no. That is the responsibility of the CF script.

src/main/scala/com/gu/kindlegen/Lambda.scala
regiskuckaertz: As you said there's only a couple hundred items, the cost of parallelism outweighs the potential speed-up. The [official recommendation](https://docs.scala-lang.org/overviews/parallel-collections/overview.html) is to use that feature at least in the several thousands elements area.

src/main/scala/com/gu/kindlegen/Lambda.scala
```diff
@@ -1,43 +1,67 @@
...
+
+        logger.log(s"Generating for $date; writing to ${customFileSettings.outputDir}")
+        val files = kindleGenerator.writeNitfBundleToDisk()
+
+        files.par.foreach { path =>
+          // TODO what if the file already exists in S3?
+          val s3path = pathPrefix + originalOutputDir.relativize(path.toAbsolutePath)
+          logger.log(s"Uploading $s3path")
+          s3.putObject(bucketName, s3path, path.toFile)
+          // TODO what if uploading to S3 fails?
```
regiskuckaertz: It can indeed [throw an exception](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#putObject-com.amazonaws.services.s3.model.PutObjectRequest-).

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: My understanding is that the official recommendation is related to computationally intensive operations, whereas this code is IO-intensive. The network latency should make up for the cost of parallelism.

On the other hand, I think that `.par` gives us _too little_ parallelism in this case, so I intend to look into using a larger thread pool.

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: No, filenames (currently) don't include a timestamp. They do contain a date (e.g. `2018-05-15`), so then two runs on the same day would cause this issue.

src/main/scala/com/gu/kindlegen/Lambda.scala
regiskuckaertz: Then I would follow AWS's recommendation to [add some randomness](https://docs.aws.amazon.com/AmazonS3/latest/dev/request-rate-perf-considerations.html) in the S3 key.

src/main/scala/com/gu/kindlegen/Lambda.scala
tomrf1: From the aws docs:
`If versioning is not enabled, this operation will overwrite an existing object with the same key`
I think this is fine, since the lambda runs daily. We can enable versioning on the s3 bucket if we change our minds

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: Do I understand correctly that you're recommending completing the process even if some operations fail?

cfn.yaml
hosamaly: Sounds good. Is there a particular key that I should use?

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: How about a simple validation like this:
```scala
require(s3.doesBucketExistV2(bucketName), s"S3 bucket $bucketName was not found")
```

src/main/scala/com/gu/kindlegen/Lambda.scala
hosamaly: I know. I'm really wondering about whether to complete the process or abort it if some files fail to be uploaded. (I should probably add some logic for retrying but the question will still remain.)

cfn.yaml
regiskuckaertz: I think we should create a new one. There is a default encryption key created for each lambda, but I haven't found a way to use it in the CF script... it must be enabled manually 😮

src/main/scala/com/gu/kindlegen/Lambda.scala
regiskuckaertz: I guess it depends on the file. If a random article (or image) fails, most people would probably be fine as they would not even read that article. Not a panacea, but a good heuristic still.

If the process is inductive (i.e. a depth-first tree traversal), then you can generate and upload the leaves and generate and upload the nodes depending on the result of the leaves.

hosamaly: I looked into doing this. The code that needs updating is in `KindleGenerator`. I found that modifying that class to ignore failures will require a considerable amount of work. Given our current priorities, I'm going to leave it until it becomes an issue.

PR #23, opened 2018-05-18
Better Publishing and Better Logging

hosamaly:
There is a lot of refactoring in this PR. Highlights include:
* Logging
* Introduction of the `Publisher` trait, making it easy to redirect output elsewhere (e.g. to SFTP).


src/main/scala/com/gu/io/Publisher.scala
```diff
@@ -0,0 +1,42 @@
...
+  /** Called before content is persisted */
+  protected def saving(key: String): Unit = { logger.trace(s"Saving $key...") }
+
+  /** Called after content is persisted */
+  protected def saved(fileName: String, link: Link): Unit = {
+    logger.trace(s"Saved $fileName as $link")
+    savedLinks.put(link, null)
+  }
+
+  protected val savedLinks: scala.collection.concurrent.Map[Link, Null] = TrieMap()
```
tomrf1: why is this a map?

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
```diff
@@ -1,116 +1,104 @@
...
+      .flatMap { articles =>
+        Future.sequence(articles.zipWithIndex.map((downloadMainImage _).tupled))
+      }
 
-    (articlesOnDisk ++ articlesOnDisk.flatMap(_.mainImage) ++ sections :+ rootManifest)
-      .map(_.link).collect {
-      case link: PathLink => link.toPath
-    }
+    Await.ready(eventualArticlesWithImages, querySettings.downloadTimeout)  // force the timeout
+      .flatMap { articlesWithImages =>
```
tomrf1: could a for/yield be neater here?

src/main/scala/com/gu/io/Publisher.scala
hosamaly: It was the only concurrent collection type in Scala that I could find!

tomrf1: ugh, that's unfortunate. Worth adding a comment here, to save the next person from wondering the same thing

hosamaly: Done.

PR #24, opened 2018-05-18
Smaller images

hosamaly:
We noticed that Amazon's preview generator takes a _long_ time when faced with master images.

This PR makes the maximum image resolution configurable.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/Image.scala
+++ b/src/main/scala/com/gu/kindlegen/Image.scala
@@ -16,15 +16,10 @@ object Image {
       e.`type` == ElementType.Image && e.relation == "main"
 
     def bestAcceptable(assets: Seq[Asset]): Option[Asset] = {
-      val assetDimensions: Seq[(Asset, Int)] = assets.map { asset =>
-        asset -> asset.typeData.flatMap { fields => Seq(fields.height, fields.width).max }
-      }.collect {
-        case (asset, Some(maxDimension)) if maxDimension <= settings.maxImageResolution =>
-          asset -> maxDimension
-      }
-
-      if (assetDimensions.isEmpty) None
-      else Some(assetDimensions.maxBy(_._2)._1)
+      assets.iterator.flatMap(MeasurableAsset.apply)
+        .filter(_.maxDimension <= settings.maxImageResolution)
+        .reduceOption(MeasurableAsset.ordering.max)
+        .map(_.asset)
     }
 
     for {
@@ -42,5 +37,15 @@ object Image {
         caption = metadata.caption,
         credit = if (metadata.displayCredit.getOrElse(true)) metadata.credit else None
       )
+  }
+
+  private class MeasurableAsset(val asset: Asset, val maxDimension: Int)
+  private object MeasurableAsset {
+    def apply(asset: Asset): Option[MeasurableAsset] =
+      asset.typeData
+        .flatMap { fields => Seq(fields.height, fields.width).max }
+        .map(new MeasurableAsset(asset, _))
+
+    def ordering: Ordering[MeasurableAsset] = Ordering.by(_.maxDimension)
   }
 }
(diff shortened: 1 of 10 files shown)


src/main/scala/com/gu/kindlegen/Image.scala
```diff
@@ -11,18 +11,27 @@ case class Image(id: String,
...
 
 object Image {
-  def mainMaster(content: Content): Option[Image] = {
+  def mainImage(content: Content, settings: QuerySettings): Option[Image] = {
     def isMainImage(e: Element): Boolean =
       e.`type` == ElementType.Image && e.relation == "main"
 
-    def isMasterAsset(a: Asset): Boolean =
-      a.typeData.flatMap(_.isMaster).getOrElse(false)
+    def bestAcceptable(assets: Seq[Asset]): Option[Asset] = {
```
regiskuckaertz: We can massage that function a little bit and come up with a more efficient version. First notice that the first expression is a fold:
```scala
val assetDimensions: Seq[(Asset, Int)] = assets.foldRight(Nil: Seq[(Asset, Int)]) { case (asset, pairs) =>
  if (asset.maxDimension <= settings.maxImageResolution)
    (asset -> asset.maxDimsension) :: pairs
  else
    pairs
}
```
(I'm assuming an implicit `maxDimension` on `Asset` that returns `Infinity` or `max(typeData.width, typeData.height)` if both are defined.) We get one intermediate list less.
 
`bestAcceptable` is the prototypical example of a `f . foldr g a` where `f` is lines 26–27. We can apply the _fusion law_ of `foldr` and find an `h` such that `f . foldr g a = foldr h b` iff

- `f a = b` (here, `None`)
- `f (g x y) = h x (f y)`

Clearly `h` has type `(Asset, Option[Asset]) => Option[Asset]`. We could calculate it but the logic is simple, so we can derive it straight away:
```scala
def bestAcceptable: Seq[Asset] => Option[Asset] =
  _.foldRight(None: Option[Asset]) { case (asset, result) =>
    if (asset.maxDimension <= settings.maxImageResolution)
      result.map(asset2 => if (asset > asset2) asset else asset2).orElse(Some(asset))
    else
      result
  }
```
(It is trivial to define an `Ordering` based on `maxDimension`.) And we get a result with no intermediate list at all 🎉

If we can assume that `assets` is a sorted sequence, we can improve that function even more. Finally, if one day you wish to have a more fine-grained definition for the size of an image, all you need to do is change `maxDimension`—nice decoupling.

tomrf1: `assets` will not be sorted as it comes back from capi, but +1 for doing this in a single pass.
Also, I'd personally always prefer `getMaxDimension(asset)` to an implicit method, but I know we all have differing views on this :)

hosamaly: I think that much of the complexity is caused by the abundance of optional values in the asset. What do you think about the following implementation?
```scala
private class MeasurableAsset(val asset: Asset, val maxDimension: Int)
private object MeasurableAsset {
  def apply(asset: Asset): Option[MeasurableAsset] =
    asset.typeData
      .flatMap { fields => Seq(fields.height, fields.width).max }
      .map(new MeasurableAsset(asset, _))

  def ordering: Ordering[MeasurableAsset] = Ordering.by(_.maxDimension)
}

def bestAcceptable(assets: Seq[Asset]): Option[Asset] = {
  assets.iterator.flatMap(MeasurableAsset.apply)
    .filter(_.maxDimension <= settings.maxImageResolution)
    .reduceOption(MeasurableAsset.ordering.max)
    .map(_.asset)
}
```

regiskuckaertz: I'm not familiar with `Iterator` but I'm assuming it's building up a computation, much like `CoYoneda`. Sweet! Whatever feels more idiomatic 🤗

mchv: Agree with @tomrf1 no implicits, @hosamaly nice implementation!

hosamaly: Thanks! 😊

PR #25, opened 2018-05-23
Better HTML => NITF Conversion

hosamaly:
This PR expands the capabilities of HTML conversion to NITF. More tags are now supported (e.g. `div`) and more tags are being filtered.


src/main/scala/com/gu/kindlegen/ArticleNITF.scala
```diff
@@ -19,7 +19,7 @@ object ArticleNITF {
   val Version = "-//IPTC//DTD NITF 3.5//EN"
   private val formatter = DateTimeFormatter.ISO_OFFSET_DATE_TIME.withZone(ZoneOffset.UTC)
 
-  private val cleaner = new Cleaner(Whitelist.relaxed  // TODO remove unwanted tags
+  private val cleaner = new Cleaner(Whitelist.relaxed
```
hosamaly: I'm a bit worried that [`Whitelist.relaxed`](https://jsoup.org/apidocs/org/jsoup/safety/Whitelist.html#relaxed--) may omit some tags unnecessarily. It doesn't include tags like `big`, `dfn`, `figure`, `mark`, `samp`, `section`, or `summary`. I think that I should add them manually. What do you think?

src/test/resources/xhtml-example.nitf
```diff
@@ -59,6 +61,13 @@
                     <p>should not be combined</p>
                     <p>with other paragraphs, not even the new wrappers</p>
                 </blockquote>
+
+                <div>
+                    A &lt;div&gt; should be converted to a &lt;block&gt;.
+                    <p>The div's contents</p>
+                    <img src="somewhere">including images</img>
+                    should be converted
+                </div>
```
hosamaly: This is a test file used in `XhtmlToNitfTransformationSpec`.

src/main/scala/com/gu/kindlegen/ArticleNITF.scala
regiskuckaertz: In the future, you might want to add `figure` (and `figcaption`) indeed.

hosamaly: I added `figure` but refrained from `figcaption` for now because I don't know how to map it to NITF.

PR #29, opened 2018-06-12
Weather support

hosamaly:
Highlights:
- Support for multiple instances of `ArticleProvider` to collect articles from various sources
- AccuWeatherClient and a weather article provider
- Combined thread pool for external HTTP calls (including CAPI, AccuWeather, and downloading images)


src/main/scala/com/gu/kindlegen/weather/WeatherSettings.scala
```diff
@@ -0,0 +1,17 @@
+package com.gu.kindlegen.weather
+
+import com.gu.kindlegen.{Image, Section}
+
+
+final case class WeatherSettings(section: Section,
+                                 minForecastsPercentage: Int,  // how many successful retrievals of forecasts to create an article
+                                 articles: Seq[WeatherArticleSettings])
```
regiskuckaertz: Are these coming from the config file? That's a bit unorthodox 😮

src/main/scala/com/gu/kindlegen/weather/accuweather/AccuWeatherClient.scala
```diff
@@ -0,0 +1,102 @@
...
+      .drop(1)  // header
+      .map(_.split(','))
+      .collect {
+        case Array(cityName, locationKey, _*) =>
+          Location(cityName, locationKey)
+      }.toSet
+  }
+
+  private[accuweather] implicit val singleForecastReader: Reader[Forecast] = (root: JValue) => {
+    val forecast = root.detect(r => (r \ "DailyForecasts") (0)) // first element of the array
```
regiskuckaertz: Isn't this a use case for `detectValue[JArray]`?

src/main/scala/com/gu/kindlegen/weather/WeatherClient.scala
```diff
@@ -0,0 +1,34 @@
...
+  case class Forecast(headline: Option[String],
+                      high: Option[Int],
+                      low: Option[Int],
+                      airQuality: Option[String],
+                      rainProbability: Option[Int]) {
+
+    val values: Iterable[Option[_]] = this.productIterator.map(_.asInstanceOf[Option[_]]).toIterable
+
+    /** Returns true if at least one field has a value */
+    def isDefined: Boolean = !isEmpty
```
regiskuckaertz: perhaps `nonEmpty` would be more accurate

hosamaly: I followed the API of `Option`, but I don't mind changing it.

src/main/scala/com/gu/kindlegen/weather/WeatherSettings.scala
hosamaly: They are defined in the configuration file [here](https://github.com/guardian/kindle-gen/pull/29/files#diff-fa2e842b741241072d2436e8ae36c3d2R51). Here is a sample from my local test config:
```
weather {
  minForecastsPercentage = 50

  section {
    id = "theguardian/mainsection/weather2"
    title = "Weather"
    link = "https://www.theguardian.com/theguardian/mainsection/weather2"
  }

  articles = [
    {
      title = "UK & Ireland Weather"
      byline = "Powered by AccuWeather"
      pageNumber = 1000  // after any other weather articles
      cities = [
        "Dublin"
        "Edinburgh"
        "London"
      ]
    }
    {
      title = "International Weather"
      byline = "Powered by AccuWeather"
      pageNumber = 1001
      cities = [
        "Cairo"
        "New York"
        "Sydney"
      ]
    }
  ]
}
```

regiskuckaertz:
I think this all makes sense 👍 when the quality tests pass

src/main/scala/com/gu/kindlegen/weather/WeatherClient.scala
```diff
@@ -0,0 +1,34 @@
...
+  final case class Location(name: String, key: String)
+
+  /** A row in our forecasts table */
+  case class Forecast(headline: Option[String],
+                      high: Option[Int],
+                      low: Option[Int],
+                      airQuality: Option[String],
+                      rainProbability: Option[Int]) {
+
+    val values: Iterable[Option[_]] = this.productIterator.map(_.asInstanceOf[Option[_]]).toIterable
```
tomrf1: I didn't know you can do this with a case class! I suppose the downside to this line is that the compiler will not complain if you add a parameter that is not an `Option`

hosamaly: Just for fun, a more complete implementation could be:
```scala
def values(p: Product): Iterable[Option[_]] = {
  p.productIterator.map {
    case o: Option[_] => o
    case t: Try[_] => t.toOption
    case e: Either[_, _] => e.toOption
    case x => Option(x)
  }
}
```

src/main/scala/com/gu/kindlegen/Settings.scala
```diff
@@ -16,15 +18,21 @@ import com.gu.io.Link.AbsoluteURL
 import com.gu.kindlegen.weather.WeatherSettings
 
 /** Encapsulates the settings of this application */
-final case class Settings(contentApi: ContentApiSettings,
+final case class Settings(credentials: Credentials,
```
regiskuckaertz: I don't think it's worth adding a wrapper if you provide accessors below

src/test/scala/com/gu/kindlegen/weather/accuweather/AccuWeatherClientSpec.scala
```diff
@@ -47,18 +52,41 @@ class AccuWeatherClientSpec extends FunSpec {
...
           .thenRespond(forecastResponse)
       }
 
       val client = AccuWeatherClient(apiKey = "", new URI("http://example.com"), downloader)
-      val actual = client.forecastFor(location).futureValue
+      val actual = client.forecastFor(cairo).futureValue
       actual shouldBe expectedForecast
     }
+
+    it("works with the actual API") {
```
regiskuckaertz: Why should a build fail if the Weather API is not working?

src/main/scala/com/gu/kindlegen/ArticlesProvider.scala
```diff
@@ -2,7 +2,20 @@ package com.gu.kindlegen
...
 
 trait ArticlesProvider {
   def fetchArticles(): Future[Seq[Article]]
 }
+
+final class CompositeArticlesProvider(providers: ArticlesProvider*) extends ArticlesProvider {
+  require(providers.nonEmpty, "providers must not be empty!")
+
+  override def fetchArticles(): Future[Seq[Article]] =
+    providers
```
regiskuckaertz: You may want to use 
```scala
Future.reduceLeft(providers.map(_.fetchArticles))(_ ++ _)
```

src/test/scala/com/gu/kindlegen/weather/accuweather/AccuWeatherClientSpec.scala
hosamaly: It's a good question, to which I don't have an answer. There are pros and cons for each side of the argument.

When I wrote this test, it caught a bug that would've caused authentication to fail. That's why I'm inclined to keep it. What do you think?

src/main/resources/reference.conf
```diff
@@ -55,6 +55,44 @@ weather {
       byline = "Powered by AccuWeather"
       pageNumber = 1000  // after any other weather articles
       cities = [
+        "Belfast"
```
regiskuckaertz: Ah now I see. That's really cool 👍

src/main/resources/reference.conf
```diff
@@ -43,10 +43,18 @@ s3 {
...
+    default {
+      id = "theguardian/mainsection/weather2"
+      title = "Weather"
+      link = "https://www.theguardian.com/theguardian/mainsection/weather2"
+    }
+
+    Sunday {
+      id = "theobserver/weather"
+      title = "Observer Programmic Weather"
+      link = "https://www.theguardian.com/theguardian/mainsection/weather2"
```
regiskuckaertz: The link is the same here. Is this an oversight?

src/test/scala/com/gu/kindlegen/weather/accuweather/AccuWeatherClientSpec.scala
regiskuckaertz: It's something you would have caught anyway and won't probably come up again. For problems with live systems I'd favour CloudWatch alarms as they're not problems with your code per se.

src/main/scala/com/gu/kindlegen/ArticlesProvider.scala
regiskuckaertz: Yeah, I don't know how it's handled internally but I assume it would do something similar.

src/test/scala/com/gu/kindlegen/weather/accuweather/AccuWeatherClientSpec.scala
LATaylor-guardian: Agreed. We don't want to couple the ability for a build to pass with an external api.

src/main/resources/reference.conf
hosamaly: There is no "Observer Weather" tag in CAPI, so I'm making this information up.

Since the purpose of the link is to point to relevant data, I thought that it made sense to link to an existing weather section.

src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
```diff
@@ -14,9 +14,12 @@ import com.gu.kindlegen._
 
 
 object GuardianArticlesProvider {
-  def apply(settings: Settings, downloader: SttpDownloader, editionDate: LocalDate)(implicit ec: ExecutionContext): GuardianArticlesProvider = {
-    val capiClient = contentApiClient(settings.contentApi, downloader)
-    new GuardianArticlesProvider(capiClient, settings.articles, editionDate)
+  def apply(contentApiSettings: ContentApiSettings,
+            capiSettings: GuardianProviderSettings,
```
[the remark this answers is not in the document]
hosamaly: You're right. How about renaming `ContentApiSettings` to `ContentApiCredentials`?

PR #31, opened 2018-06-21
Combine and re-order sections

hosamaly:
Editorial would like the ability to combine sections together and reorder them at will. This PR enables this via a configuration file.


src/main/resources/guardian-sections.conf
```diff
@@ -0,0 +1,94 @@
+mainSections = [
```
hosamaly: Based on the data [in this document](https://docs.google.com/spreadsheets/d/1kawGUhjh4LdeYzXuecUotjwv51_Qnx4KsuWHxV7j8Yw/edit?usp=drive_web&ouid=102546719289671538304)

hosamaly:
While reviewing, please note that the last commit changes much of the logic in the commit before it. You may want to review the final result.

PR #32, opened 2018-06-22
Fix a bug where spaces around link anchors were lost

hosamaly:
As an example, the following HTML:
```html
<p>Orbital play <a>Brighton Racecourse</a> on 29 June and <a>Zebedee’s Yard, Hull,</a> on 30 June. <a>Then touring</a> until 10 August.</p>
```
used to result in:
```html
<p>Orbital play BrightonRacecourseon 29 June andZebedee’s Yard, Hull, on 30 June.Then touringuntil 10 August.</p>
```
This was caused by the call to `Utility.trim`. We used to trim the input so that we won't wrap the spaces between XML tags (e.g. `<body> <body.content>`) in paragraphs. This commit fixes the logic that would have caused the incorrect wrapping in the first place.

### DIFF
--- a/src/main/scala/com/gu/kpp/nitf/XhtmlToNitfTransformer.scala
+++ b/src/main/scala/com/gu/kpp/nitf/XhtmlToNitfTransformer.scala
@@ -114,6 +114,7 @@ class XhtmlToNitfTransformer(config: HtmlToNitfConfig) {
   private val enrichedTextParentTags = config.nitf.enrichedTextParentTags
   private def enrichedTextNode(node: Node) = config.nitf.enrichedTextTags.contains(node.label)
   private def nonEnrichedTextNode(node: Node) = !enrichedTextNode(node)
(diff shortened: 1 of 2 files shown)


LATaylor-guardian:
👍

src/main/scala/com/gu/kpp/nitf/XhtmlToNitfTransformer.scala
```diff
@@ -20,7 +20,7 @@ class XhtmlToNitfTransformer(config: HtmlToNitfConfig) {
   }
 
   // the order of these rules is important
-  private def transformationRules = Seq(
+  private lazy val transformationRules = Seq(
```
tomrf1: why is this lazy?

tomrf1: Let's avoid lazy unless it's essential

hosamaly: Done

PR #34, opened 2018-06-25
Cartoons and Bylines

hosamaly:
* Use trail text for cartoon captions if a caption isn't provided
* Set the byline of all articles, falling back to a dash. The byline is required. If it's empty then Kindle will show "No byline".

### DIFF
--- a/cfn.yaml
+++ b/cfn.yaml
@@ -52,7 +52,7 @@ Resources:
       FunctionName: !Sub ${App}-${Stage}
       Description: Converting content to NITF format
       Runtime: java8
-      Handler: com.gu.kindlegen.Lambda::handler
+      Handler: com.gu.kindlegen.app.Lambda::handler
       MemorySize: 3008  # We usually use < 300 MB of RAM. However, we're I/O-bound so increasing cores improves speed drastically.
       Timeout: 300
       Tracing: Active
--- a/src/main/scala/com/gu/kindlegen/app/Settings.scala
+++ b/src/main/scala/com/gu/kindlegen/app/Settings.scala
@@ -113,7 +113,7 @@ object WeatherSectionsReader extends ValueReader[Map[DayOfWeek, Section]] {
 
     val dailySections = config.as[Map[String, Section]](path)
       .map {
-        case (dayName, section) => Try(DayOfWeek.valueOf(dayName)).toOption -> section
+        case (dayName, section) => Try(DayOfWeek.valueOf(dayName.toUpperCase)).toOption -> section
       }
 
     val defaultSection = dailySections(None)
--- a/src/main/scala/com/gu/kindlegen/app/Settings.scala
+++ b/src/main/scala/com/gu/kindlegen/app/Settings.scala
@@ -12,7 +12,7 @@ import net.ceedubs.ficus.readers.ValueReader
 import net.ceedubs.ficus.readers.ArbitraryTypeReader._
 
 import com.gu.config._
-import com.gu.contentapi.client.model.v1.TagType
+import com.gu.contentapi.client.model.v1.{Tag, TagType}
 import com.gu.io.Link
 import com.gu.io.Link.AbsoluteURL
 import com.gu.kindlegen._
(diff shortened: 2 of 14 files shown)


src/main/scala/com/gu/kindlegen/Article.scala
```diff
@@ -46,18 +48,28 @@ object Article {
             fields: ContentFields,
             sectionTag: Tag,
             settings: GuardianProviderSettings): Article = {
+    val bodyBlocks = getBodyBlocks(content).map(_.trim)
+
+    val isCartoon = sectionTag.id.contains("/cartoon")
```
hosamaly: @LATaylor-guardian I left this generic to capture any cartoons. Are you satisfied with this logic, or would you rather check for the two cartoon sections by name?

tomrf1: This _should_ be ok as it captures both `type/cartoon` and `tone/cartoons`, but I can find examples of cartoons without either of these tags:
https://content.guardianapis.com/search?tag=commentisfree/series/observer-comment-cartoon,-type/cartoon,-tone/cartoons
https://content.guardianapis.com/search?tag=commentisfree/series/guardian-comment-cartoon,-type/cartoon,-tone/cartoons

Really this is a failure in how we tag content, but it's easier to just match against these 2 series tags here as well

LATaylor-guardian: > but it's easier to just match against these 2 series tags here as well
Agreed.

hosamaly: Which approach do you think is best? I'd vote for option 3 if you agree that the probability of a false positive is very low.
```scala
// option 1: very specific
val cartoonTags = Set(
      "type/cartoon", "tone/cartoons",
      "commentisfree/series/guardian-comment-cartoon",
      "commentisfree/series/observer-comment-cartoon")
cartoonTags.contains(section.id)

// option 2: somewhat flexible
sectionTag.id.contains("/cartoon") || sectionTag.id.contains("-cartoon")

// option 3: most flexible
sectionTag.id.matches(raw".*\bcartoons?\b.*")
```

hosamaly: I noticed today that the section tag (of type `NewspaperBookSection`) isn't the right one to check. So, the correct options are:
```scala
// option 1: ask and check for specific tags
PrintSentSearchQuery.showTags("type", "tone", "series")
val cartoonTagIds = Set("type/cartoon", "tone/cartoons", "commentisfree/series/guardian-comment-cartoon", "commentisfree/series/observer-comment-cartoon")
val isCartoon = content.tags.exists(cartoonTagIds)

// option 2: ask for all tags and check for the common pattern
PrintSentSearchQuery.showTags("all")
val isCartoon = content.tags.exists(tag => tag.id.contains("cartoon") || tag.id.contains("-cartoon"))

// option 3: ask for all tags and check for the word "cartoon"
PrintSentSearchQuery.showTags("all")
val isCartoon = content.tags.exists(tag => tag.id.matches(raw".*\bcartoon\b.*"))
```
Which one do you think is preferable?

tomrf1: Now that we know exactly which tags to look for, let's use option 1, to avoid picking up anything else by mistake in the future

cfn.yaml
```diff
@@ -52,7 +52,7 @@ Resources:
       FunctionName: !Sub ${App}-${Stage}
       Description: Converting content to NITF format
       Runtime: java8
-      Handler: com.gu.kindlegen.Lambda::handler
+      Handler: com.gu.kindlegen.app.Lambda::handler
```
hosamaly: I had forgotten to do this in PR #35.

src/main/scala/com/gu/kindlegen/app/Settings.scala
```diff
@@ -87,13 +83,37 @@ object S3SettingsReader extends ValueReader[S3Settings] {
...
+}
+
 object WeatherSectionsReader extends ValueReader[Map[DayOfWeek, Section]] {
   override def read(config: Config, path: String): Map[DayOfWeek, Section] = {
     implicit val linkReader = AbsoluteURLReader
 
     val dailySections = config.as[Map[String, Section]](path)
       .map {
-        case (dayName, section) => Try(DayOfWeek.valueOf(dayName)).toOption -> section
+        case (dayName, section) => Try(DayOfWeek.valueOf(dayName.toUpperCase)).toOption -> section
```
hosamaly: This is a fix for a regression bug.

src/main/scala/com/gu/kindlegen/capi/KindlePublishingSearchQuery.scala
```diff
@@ -1,42 +1,42 @@
...
-    val endOfDay = startOfDay.plus(1, DAYS).minusNanos(1)
+
+    val newspaperEditionDate = date.atStartOfDay.toInstant(ZoneOffset.UTC)
 
     PrintSentSearchQuery()
       .tag(combineParams(WhiteListedTags ++ negate(BlackListedTags)))
-      .fromDate(startOfDay)
-      .toDate(endOfDay)
+      .fromDate(newspaperEditionDate)
+      .toDate(newspaperEditionDate)
```
regiskuckaertz: `from == to`?

regiskuckaertz: Hmm, another instance demonstrating `to` should be exclusive

hosamaly: The `newspaperEditionDate` is a date, not a timestamp. It's always like `2018-07-02T00:00:00Z` (even in the Summer!).
I used to give the whole day as a range, but I got worried that changes to the field to support BST may result in incorrect results, so I used the value that is guaranteed to work correctly (or not to work at all).

PR #36, opened 2018-07-02
Create a stable S3 path to the current issue

hosamaly:
We configure Amazon's KPP with a stable URL that should point to each day's data.
This commit enables us to have a stable path within the S3 bucket via a redirect.

The redirect is recreated every time the Lambda is run to point to the generated
files. For example, the following link:
http://my-bucket.s3-website-eu-west-1.amazonaws.com/CODE/CurrentIssue/hierarchical-title-manifest.xml
will point to:
http://my-bucket.s3-website-eu-west-1.amazonaws.com/CODE/2018-07-02/hierarchical-title-manifest.xml

### DIFF
--- a/src/main/scala/com/gu/kindlegen/app/Lambda.scala
+++ b/src/main/scala/com/gu/kindlegen/app/Lambda.scala
@@ -6,8 +6,9 @@ import java.time.ZoneOffset.UTC
 import java.time.temporal.ChronoUnit.HOURS
 
 import scala.collection.JavaConverters._
-import scala.concurrent.{Await, ExecutionContext}
+import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.concurrent.duration._
+import scala.util.Failure
 
 import com.amazonaws.services.lambda.runtime.Context
 import com.amazonaws.services.s3.AmazonS3ClientBuilder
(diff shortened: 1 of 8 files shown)


src/main/scala/com/gu/kindlegen/app/Lambda.scala
```diff
@@ -113,22 +117,41 @@ class Lambda(settings: Settings, date: LocalDate) extends Logging {
   def doRun(remainingTimeInMillis: => Long): Unit = {
     import scala.concurrent.ExecutionContext.Implicits.global
 
+    val publisher = s3Publisher(settings)
+    logger.info(s"Starting to publish files for $date; uploading to s3://${settings.s3.absolutePath.source}")
+
+    val deletePublicDir = publisher
```
regiskuckaertz: Can you run actions in parallel?

src/main/scala/com/gu/kindlegen/app/Lambda.scala
```diff
@@ -113,22 +117,41 @@ class Lambda(settings: Settings, date: LocalDate) extends Logging {
...
+      .undirect(settings.s3.publicDirectory)
+      .andThen { case Failure(error) => logger.warn("Deleting the public directory failed!", error) }
+
+    val generateFiles = deletePublicDir
```
regiskuckaertz: I don't think `generateFiles` should wait on `deletePublicDir`, but `publishPublicly` should. wdyt?

src/main/scala/com/gu/kindlegen/app/Lambda.scala
hosamaly: We want to make sure that Amazon won't be seeing partial content, so we want the operation to fail if this step fails. This is what we agreed upon, right @tomrf1?

regiskuckaertz: Oh I see 👍

PR #39, opened 2018-07-03
Give generated weather articles the same pubDate as other articles

hosamaly:
It appears that having different publication dates causes an issue when
Amazon KPP generates an edition. In particular, KPP generates an edition
that only contains the sections with the latest publication date.

This isn't a problem in the live code, but it becomes an issue when we
try to preview older editions. It would also become a problem if the
edition was being generated at 11:59 PM on the day before publication.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/Manifests.scala
+++ b/src/main/scala/com/gu/kindlegen/Manifests.scala
@@ -6,13 +6,13 @@ import com.gu.io.Link
 
 
 object ArticlesManifest {
-  def apply(section: BookSection, buildInstant: Instant = Instant.now): RssManifest = {
+  def apply(book: BookSection, buildInstant: Instant = Instant.now): RssManifest = {
     RssManifest(
-      title = section.title,
-      link = section.link,
+      title = book.title,
+      link = book.link,
       buildInstant = buildInstant,
-      publicationDate = section.publicationDate,
-      items = section.articles.map(article => RssItem(article.title, article.link))
+      publicationDate = book.publicationDate,
+      items = book.articles.map(article => RssItem(article.title, article.link))
     )
   }
 }
--- a/src/main/scala/com/gu/kindlegen/app/Lambda.scala
+++ b/src/main/scala/com/gu/kindlegen/app/Lambda.scala
@@ -1,13 +1,13 @@
 package com.gu.kindlegen.app
 
 import java.nio.file.{Files, Path}
-import java.time.{Instant, LocalDate, LocalTime, ZoneId}
+import java.time._
 import java.time.ZoneOffset.UTC
 import java.time.temporal.ChronoUnit.HOURS
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{Await, ExecutionContext, Future}
-import scala.concurrent.duration._
+import scala.concurrent.duration.{Duration, _}
 import scala.util.Failure
 
 import com.amazonaws.services.lambda.runtime.Context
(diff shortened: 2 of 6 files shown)


hosamaly:
This issue makes me think whether we should use articles' `newspaperEditionDate`. I'm guessing that it may be safer to overwrite the publication date of all articles when we generate the NITF. What do you think?

src/main/scala/com/gu/kindlegen/Manifests.scala
```diff
@@ -6,13 +6,13 @@ import com.gu.io.Link
 
 
 object ArticlesManifest {
-  def apply(section: BookSection, buildInstant: Instant = Instant.now): RssManifest = {
+  def apply(book: BookSection, buildInstant: Instant = Instant.now): RssManifest = {
```
hosamaly: Just a rename

src/main/scala/com/gu/kindlegen/app/Lambda.scala
```diff
@@ -122,7 +122,7 @@ class Lambda(settings: Settings, date: LocalDate) extends Logging {
 
     val deletePublicDir = publisher
       .undirect(settings.s3.publicDirectory)
-      .andThen { case Failure(error) => logger.warn("Deleting the public directory failed!", error) }
+      .andThen { case Failure(error) => logger.error("Deleting the public directory failed!", error) }
```
hosamaly: Side fix: since the future and the subsequent futures will fail, we should log an error. It should be changed back to `warn` if the error is ignored.

src/main/scala/com/gu/kindlegen/app/Lambda.scala
```diff
@@ -171,7 +171,8 @@ class Lambda(settings: Settings, date: LocalDate) extends Logging {
                              (implicit ec: ExecutionContext): ArticlesProvider = {
     val client = AccuWeatherClient(settings.accuWeather, downloader)
     val section = settings.weather.sections(date.getDayOfWeek)
-    new DailyWeatherForecastProvider(client, section, settings.weather)
+    val editionDate = date.atStartOfDay.atOffset(ZoneOffset.UTC)  // same offset as CAPI articles
+    new DailyWeatherForecastProvider(client, section, editionDate, settings.weather)
```
hosamaly: This is the actual fix.

src/test/scala/com/gu/kindlegen/KindleGeneratorSpec.scala
```diff
@@ -41,7 +42,8 @@ class KindleGeneratorSpec extends FunSpec with TempFiles {
     val weatherClient = AccuWeatherClient(settings.accuWeather.apiKey, settings.accuWeather.baseUrl, downloader)
     val publisher = FilePublisher(fileSettings.outputDir.resolve(editionDate.toString))
     val capiProvider = GuardianArticlesProvider(settings.contentApi, settings.articles, downloader, editionDate)
-    val weatherProvider = new DailyWeatherForecastProvider(weatherClient, settings.weather.sections(editionDate.getDayOfWeek), settings.weather)
+    val weatherProvider = new DailyWeatherForecastProvider(
+      weatherClient, settings.weather.sections(editionDate.getDayOfWeek), editionDate.atStartOfDay.atOffset(UTC), settings.weather)
```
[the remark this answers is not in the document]
hosamaly: Doesn't the name of the method `atOffset` provide enough context?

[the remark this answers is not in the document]
regiskuckaertz: I mean, as a general rule, it is best being consistent as you've used `ZoneOffset.UTC` elsewhere. Other cases might not be as obvious, e.g. `ElementType.Tweet`. `Tweet` alone might be data or metadata, the prefix removes that confusion. Agreed that in this case it is not that big of a deal.

regiskuckaertz:
> I'm guessing that it may be safer to overwrite the publication date of all articles when we generate the NITF.

I think it is perfectly acceptable and if it makes the behaviour of KPP more consistent then by all means 🚢 It should be equal to `editionDate` anyway (modulo time set to 00:00:00), no?

PR #40, opened 2018-07-04
Alarms

hosamaly:
* Raise an alarm if the lambda is not invoked within 1 day and 5 minutes of the last invocation
* Raise an alarm if the lambda invocation results in an error
* Raise an alarm if the logs show a warning or error

### DIFF
--- a/cfn.yaml
+++ b/cfn.yaml
@@ -57,7 +57,7 @@ Resources:
   Lambda:
     Type: AWS::Serverless::Function
     Properties:
-      FunctionName: !Sub ${App}-${Stage}  # must be kept in sync with FunctionName everywhere in the file
+      FunctionName: !Sub ${App}-${Stage}  # must be kept in sync with FunctionName and LogGroupName everywhere in the file
       Description: Converting content to NITF format
       Runtime: java8
       Handler: com.gu.kindlegen.app.Lambda::handler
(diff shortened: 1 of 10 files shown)


cfn.yaml
```diff
@@ -106,3 +110,66 @@ Resources:
               - s3:GetBucketWebsite
               - s3:PutBucketWebsite
             Resource: !Sub "arn:aws:s3:::${OutputBucket}"
+
+  LambdaLogErrors:
+    Type: "AWS::Logs::MetricFilter"
```
tomrf1: 💯

hosamaly:
One side effect of having these independent alarms is that a lambda failure will, most probably, set off two alarms: one for the lambda's error and another for a log error.

tomrf1:
@hosamaly true. Does this mean the `LambdaErrorAlarm` is redundant, as `LambdaLogErrorAlarm` will always cover the same incidents?

hosamaly:
I wouldn't go as far as saying that it's redundant. Any errors that are not captured in our logic would not make it to the logs. Generally speaking, any `Try` or `Future` transformation that results in a fatal (i.e. non-`NonFatal`) exception wouldn't be captured in our logs, but `LambdaErrorAlarm` should capture them.

This reminds me that we may want to customise the execution context to make sure that it captures such errors and logs them appropriately.

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
```diff
@@ -108,7 +115,54 @@ class KindleGenerator(provider: ArticlesProvider,
...
+    if (xsdSources.nonEmpty) {
+      try {
+        Some(xmlSchema(xsdSources))
+      } catch {
+        case e: Throwable =>
+          logger.warn(s"Couldn't create validation schema for $name!", e)
+          None
+      }
+    } else {
+      logger.warn(s"Couldn't load schema files for $name!")
```
hosamaly: We seem to hit this line in AWS, but not locally! I'll try to find out why...

hosamaly: Fixed in 3fc7afc

PR #42, opened 2018-07-06
Pre-run the Lambda to detect issues early

hosamaly:
I was asked to configure a "test run" for the lambda around midnight so that we can detect issues earlier than the final run around 1 AM.

### DIFF
--- a/cfn.yaml
+++ b/cfn.yaml
@@ -85,14 +85,18 @@ Resources:
                 ${CustomConfig}
               }
       Events:
-        DailyGMT:
+        DailyGMTRun:
           Type: Schedule
           Properties:
             Schedule: cron(10 1 * * ? *)  # Run at 01:10 AM GMT (UTC+0) every day
-        DailyBST:
+        DailyBSTRunGMTPrerun:
           Type: Schedule
           Properties:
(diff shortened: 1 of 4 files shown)


hosamaly:
With this PR, some expectations have changed.
1. The Lambda runs at 12:10 AM. We call it a "pre-run". Nonetheless, it's a full run.
2. The Lambda runs again at 01:10 AM. It creates a new directory and publishes new files. It may include more data than the "pre-run".
3. An alarm will be raised if either of the runs fails (or prints a warning to the log).
4. The `LambdaNotInvokedAlarm` will _not_ be raised unless _both_ runs are skipped (e.g. if the deploy bucket is inaccessible).
5. If the second run fails to be invoked (or fails to parse settings), then the `CurrentIssue` redirect will be pointing to the pre-run. If it doesn't run at all then we won't be notified!
5. If the second run fails, then the `CurrentIssue` redirect would have been removed. If we wish to use the results of the pre-run (assuming it was successful), then we need to either create the redirect manually or update the URL in Amazon's KPP Console.

tomrf1:
@hosamaly thanks for clarifying

PR #46, opened 2018-07-17
Upload zip archive to S3

hosamaly:
Archive the full contents of the publication into a single archive for easy download.


src/main/scala/com/gu/io/aws/S3Publisher.scala
```diff
@@ -1,53 +1,67 @@
...
+    tmpPublisher.zipPublications(archiveName).map(persist(archiveName, _))
+  }
 
-      RelativePath.from(fileName, relativeTo = settings.absolutePath)
-    }.andThen { case _ =>
-      tmpPublisher.delete(fileName)
-    }
+  override def close(): Unit = {
+    val deleteAll = tmpPublisher.publications.map(link => tmpPublisher.delete(link.source))
+    Await.ready(Future.sequence(deleteAll), 1.minute)  // ignore errors
```
regiskuckaertz: tip: you can just use `Future.sequence(deleteAll).ready(1.minute)`

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
```diff
@@ -33,11 +33,13 @@ class KindleGenerator(provider: ArticlesProvider,
 
   def fetchArticles(): Future[Seq[Article]] = {
     provider.fetchArticles()
-      .map { results =>
-        val minArticles = publishingSettings.minArticlesPerEdition
-        require(results.length >= minArticles,
-          s"Not enough articles to generate an edition! Expected >= $minArticles, Found ${results.length}")
-        results
+      .flatMap { results =>
```
regiskuckaertz: If the error message is not that big of a deal, you can just filter:
```scala
provider.fetchArticles().filter(rs => rs.length >= publishingSettings.minArticlesPerEdition)
```

src/main/scala/com/gu/io/aws/S3Publisher.scala
hosamaly: `Await.ready` is slightly different in that it uses [`scala.concurrent.blocking`](https://github.com/scala/scala/blob/v2.12.6/src/library/scala/concurrent/package.scala#L131), which is...
> Used to designate a piece of code which potentially blocks, allowing the current `BlockContext` to adjust the runtime's behavior.
>  Properly marking blocking code may improve performance or avoid deadlocks.

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
hosamaly: It's not just an error message. It's a failed future, which will eventually cause the Lambda to exit with an error, stopping publishing and raising an alarm.

regiskuckaertz: Precisely. Then `IllegalArgumentException` or `NoSuchElementException` doesn't make any difference. In both cases, we're stretching their meaning anyway. What I'm more concerned about is the semantics of the operation: none of them require the full machinery of a future, which is why `flatMap` is over the top. You are actually working on the underlying `Try[A]`, and the API gives you the ability to expose that intent so that the future you or any other developer can look at this code and understand its meaning at a glimpse.

regiskuckaertz: If you absolutely need a bespoke error then you can create one:
```scala
provider.fetchArticles().transform {
  case Success(rs) if rs.length < publishingSettings.minArticlesPerEdition => Failure(new NotEnoughArticlesException(rs.length, publishingSettings.minArticlesPerEdition)
  case x => x
}
```

src/main/scala/com/gu/io/aws/S3Publisher.scala
```diff
@@ -1,53 +1,67 @@
...
 
-  override def persist(content: Array[Byte], fileName: String): Future[Link] = {
+  override def persist(content: Array[Byte], fileName: String): Future[PublishedLink] = {
     // write the file to disk temporarily; the S3 client API infers the metadata from files automatically
-    tmpPublisher.persist(content, fileName).map { relativePath =>
-      val s3Path = normalizeDir(bucketDirectory) + fileName
+    tmpPublisher.persist(content, fileName).map(persist(fileName, _))
+  }
+
+  private def persist(fileName: String, relativePath: RelativePath) = {
```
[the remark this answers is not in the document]
hosamaly: Could you please elaborate on this difference between `map` and `flatMap`?

I can see that [`Future#map`](https://github.com/scala/scala/blob/v2.12.6/src/library/scala/concurrent/Future.scala#L288) uses [`DefaultPromise#transform`](https://github.com/scala/scala/blob/v2.12.6/src/library/scala/concurrent/impl/Promise.scala#L27), and [`Future#flatMap`](https://github.com/scala/scala/blob/v2.12.6/src/library/scala/concurrent/Future.scala#L302) uses [`DefaultPromise#transformWith`](https://github.com/scala/scala/blob/v2.12.6/src/library/scala/concurrent/impl/Promise.scala#L34), both of which create chain promises `onComplete`. Can you help me spot the difference, please?

src/main/scala/com/gu/io/aws/S3Publisher.scala
hosamaly: I wonder why the method isn't package-private.

src/main/scala/com/gu/kindlegen/KindleGenerator.scala
hosamaly: Alright. I've switched to the use of `transform`.

I don't (currently) need a bespoke error because there is no special handling for this case.

src/main/scala/com/gu/io/aws/S3Publisher.scala
regiskuckaertz: Yeah, it's almost as if they wanted you to make the mistake 😭

src/main/scala/com/gu/io/aws/S3Publisher.scala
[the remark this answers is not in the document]
regiskuckaertz: Yes, the difference lies in where the callback is linked (in the original promise with `flatMap`). There is some detail around that https://www.beyondthelines.net/computing/scala-future-and-execution-context/:

> However for `flatMap` (which is implemented using `transformWith`) things are more complicated. Like `transform`, `transformWith` creates a new Promise `p`. But this time the callback also returns a Future which might as well be a DefaultPromise. This new Promise can only complete when the promise `p` does. Therefore it’s safe to “link” this promise to `p` and add all of its callbacks to the callbacks of `p`. It’s like if the whole promise chain gets “collapsed” onto the root promise. This is a very useful thing to do as it may avoid some memory links.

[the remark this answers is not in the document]
hosamaly: We discussed this offline. As the method is private and self-contained (i.e. doesn't create more futures), it's fine to leave it as-is.

PR #47, opened 2018-07-17
Compact weather tables

hosamaly:
Custom city names and custom weather headlines


src/main/scala/com/gu/kindlegen/accuweather/AccuWeatherClient.scala
```diff
@@ -26,17 +26,25 @@ object AccuWeatherClient {
...
             apiBaseUri: URI,
             downloader: SttpDownloader)(implicit ec: ExecutionContext): AccuWeatherClient = {
-    new AccuWeatherClient(apiKey, apiBaseUri, knownLocations, downloader)
+    new AccuWeatherClient(apiKey, apiBaseUri, location, downloader)
   }
 
+  def location(cityName: String): Option[Location] =
+    knownLocationsMap.get(cityName.toLowerCase)
+
+  lazy val knownLocationsMap = knownLocations
```
regiskuckaertz: Could you please add a type ascription? I assume `Map[Set[String], Location]`

hosamaly: Done

PR #49, opened 2018-12-03
Filter out potential duplicates

regiskuckaertz:
A user complained about articles sometimes appearing twice in the same edition:

> Firstly may I refer you to my email of the 18th November. It's happened again!  The article by Arwa Mahdawin("*Only a fool would have bought into bit coin...*") is in the Kindle *twice*, on the same day (today, 26/11), I suggest strongly that you take action against the subeditor culprit responsible for this sloppy output. I do pay for this edition although I'm aware that I could read many of the articles for free, on line.

This changes takes them out at the source.

### DIFF
--- a/src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
+++ b/src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
@@ -50,12 +50,7 @@ class GuardianArticlesProvider(capiClient: ContentApiClient,
   }
 
   def articles(results: Seq[Content]): Seq[Article] = {
-    val (_, articles) = results.foldLeft[(Set[String], Seq[Article])]((Set.empty, Nil)) { case ((ids, articles), article) =>
-      if (!ids(article.id))
-        ids -> (articles ++ toArticle(article))
-      else
-        (ids + article.id) -> articles
-    }
+    val articles = results.groupBy(_.id).flatMap(_._2.headOption).toSeq.flatMap(toArticle)
     logger.info(s"Processed ${articles.length} articles.")
     articles
   }
--- a/src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
+++ b/src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
@@ -50,7 +50,12 @@ class GuardianArticlesProvider(capiClient: ContentApiClient,
   }
 
   def articles(results: Seq[Content]): Seq[Article] = {
-    val articles = results.flatMap(toArticle)
+    val (_, articles) = results.foldLeft[(Set[String], Seq[Article])]((Set.empty, Nil)) { case ((ids, articles), article) =>
+      if (!ids(article.id))
+        ids -> (articles ++ toArticle(article))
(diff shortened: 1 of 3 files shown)


regiskuckaertz:
☝️ why don't TeamCity results show up there?

LATaylor-guardian:
> 👇 why don't TeamCity results show up there?

Probably needs the "Github report status" plugin configured in TC.

regiskuckaertz:
@LATaylor-guardian could you make me admin in the Off-platforms groups on TC?

src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
```diff
@@ -50,7 +50,12 @@ class GuardianArticlesProvider(capiClient: ContentApiClient,
   }
 
   def articles(results: Seq[Content]): Seq[Article] = {
-    val articles = results.flatMap(toArticle)
+    val (_, articles) = results.foldLeft[(Set[String], Seq[Article])]((Set.empty, Nil)) { case ((ids, articles), article) =>
```
LATaylor-guardian: Can you write this more simply as something like:

```
articles.groupBy(_.id).flatMap(_._2.headOption).toSeq

```

regiskuckaertz: Sure!

LATaylor-guardian:
👍

regiskuckaertz:
Thanks. Hmm, it looks correct. Perhaps the token is invalid then 🤔

regiskuckaertz:
Can you use the GH approve UI please?

mchv:
@regiskuckaertz May you let user help knows that the issue has now been fixed? 🙏

regiskuckaertz:
@mchv Yep I sent them an email yesterday ✌️

src/main/scala/com/gu/kindlegen/capi/GuardianArticlesProvider.scala
```diff
@@ -50,7 +50,7 @@ class GuardianArticlesProvider(capiClient: ContentApiClient,
   }
 
   def articles(results: Seq[Content]): Seq[Article] = {
-    val articles = results.flatMap(toArticle)
+    val articles = results.groupBy(_.id).flatMap(_._2.headOption).toSeq.flatMap(toArticle)
```
hosamaly: It may be slightly more efficient to use `flatMap(toArticle)(scala.collection.breakOut)` than `.toSeq.flatMap(toArticle)`. :)

regiskuckaertz: Good point!

regiskuckaertz: Oh, apparently it's going out in 2.13 in favour of views with the new collections library, something to keep in mind!

hosamaly: In 2.13,we should be able to use
`results.distinctBy(_.id).flatMap(toArticle)`

PR #51, opened 2019-01-16
Remove duplicates

regiskuckaertz:
There are three endpoints for print content:

- /print-sent
- /print-sent/live
- /print-sent/draft

The first one returns both draft and live content, which results in draft content being published in Kindle editions. This doesn't seem to appear often, but it does, and is especially noticeable since there appears to be journalists who write a draft story and then publish the live version in a separate article.

The initial fix was completely unnecessary so I removed it.


mchv:
This is certainly not the correct fix, as draft content is expected to be published in the Kindle

regiskuckaertz:
😲 draft content? ok let's chat tomorrow... it's frustrating this isn't documented anywhere and completely against everything else we do.

PR #52, opened 2019-02-06
Fix incorrect location keys

regiskuckaertz:
A few cities were wrongly identified. I checked using their API.


NataliaLKB:
Nice find @regiskuckaertz !

regiskuckaertz:
[Wrong weather forecasts](https://trello.com/c/Revk2I5V/595-wrong-weather-forecasts)

PR #54, opened 2019-05-15
Bump thrift-related deps

regiskuckaertz:
This one is more involved because things were failing at runtime. As it turns out, it's not a good idea to reuse the package structure of dependencies. Even though compilation was fine, I got `NoClassDefFound` exceptions. Why it appeared now and not before, I still haven't fully understood.


regiskuckaertz:
Tested on CODE

hosamaly:
I'm surprised that you got those `NoClassDefFound` exceptions! Thanks for fixing it!

PR #55, opened 2019-08-16
Handle Daylight Saving Time

regiskuckaertz:
The lambda will run 3x whenever DST changes and this created problems in the past. This is addressed here by comparing the local time at which the lambda is supposed to run with the universal time at which it effectively runs.


src/test/scala/com/gu/kindlegen/app/LambdaSpec.scala
```diff
@@ -0,0 +1,51 @@
+package com.gu.kindlegen.app
+
+import org.scalatest.{Assertion, FunSpec}
+import org.scalatest.Inspectors._
+import org.scalatest.Matchers._
+import java.time._
+
+class LambdaSpec extends FunSpec {
```
tomrf1: 💯

src/test/scala/com/gu/kindlegen/app/LambdaSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 2, 17, 0, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 2, 17, 1, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
+    }
+  }
+
+  describe("Lambda.timeToRun - no DST -> DST") {
+    it("only runs twice on the day DST starts") {
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 3, 30, 23, 0, 0, 1, ZoneOffset.UTC)) shouldBe false
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 3, 31, 0, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 3, 31, 1, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
```
JustinPinner: If this works then I _think_ it's reliable. 01:00 doesn't actually exist on this date does it? I think it jumps immediately from 01:00 to 02:00.

regiskuckaertz: It doesn't exist in London time indeed!

src/test/scala/com/gu/kindlegen/app/LambdaSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 8, 17, 0, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 8, 17, 1, 0, 0, 1, ZoneOffset.UTC)) shouldBe false
+    }
+  }
+
+  describe("Lambda.timeToRun - DST -> no DST") {
+    it("only runs twice on the day DST stops") {
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 10, 26, 23, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 10, 27, 0, 0, 0, 1, ZoneOffset.UTC)) shouldBe true
+      Lambda.isTimeToRun(runSettings, ZonedDateTime.of(2019, 10, 27, 1, 0, 0, 1, ZoneOffset.UTC)) shouldBe false
```
JustinPinner: Going back from BST to GMT happens at 02:00 and means we get two 01:00s I think. Might this run again then? Perhaps we need to check `2019, 10, 27, 2, 0, 0, 1`

regiskuckaertz: Fortunately we don't have to, because the lambda is triggered 3 times in the UTC time zone. (23pm, 12am and 1am). All we need to check is which two of these match 12 am and 1am in London time.

PR #56, opened 2021-06-21
Add ability to run kindle-gen locally

Fweddi:
## What does this change?
This PR allows a developer to run the kindle-gen lambda locally, provided they use the correct configuration settings.

## How to test
Follow the instructions in [the README](https://github.com/guardian/kindle-gen/blob/fp/local-kindle-gen/README.md#running-locally) so that you are using the right configuration settings and local environment.

Use `sbt run` to run the lambda, and inspect the CODE S3 bucket to see the uploaded output.

## How can we measure success?
The lambda should execute without error, and upload correctly formatted kindle data to the DEV folder in the CODE S3 bucket.

## Have we considered potential risks?
Pointing to the production bucket could risk the Kindle publishing unintended data, so be sure to follow the instructions in the README closely.

Excessive use of keys used in production (e.g. Accuweather has the same key for PROD and CODE, which I used locally) could cause us to reach the rate limit for that key. Ideally, we would have separate keys for different environments, but so far I have been unable to track down who maintains our relationship with Accuweather (this service is used by many applications in different teams).


src/main/scala/com/gu/kindlegen/app/Settings.scala
```diff
@@ -78,13 +78,15 @@ private object AbsoluteURLReader extends ValueReader[Link] {
...
+    val environment     = config.as[String](Environment).stripSuffix("/")
     val bucketDirectory = config.as[String](BucketDirectory).stripSuffix("/")
     val publicDirectory = config.as[String](PublicDirectory).stripSuffix("/")
     val tmpDirOnDisk    = config.as[Option[Path]](TmpDirOnDisk)
-    S3Settings(bucketName, bucketDirectory, publicDirectory, tmpDirOnDisk)
+    S3Settings(bucketName, environment, bucketDirectory, publicDirectory, tmpDirOnDisk)
   }
 
   private final val BucketName = "bucket"
+  private final val Environment = "prefix"
```
Fweddi, line 89: I've elected to use the S3 folder prefix to define the environment, as (a) this variable needs to be set to PROD / CODE / DEV, and (b) it eliminates the need to add another environment variable to our build.

Am happy to add a new config variable if it's desirable.

JustinPinner (changes requested):
I wonder if most of this can be done without changing the main lambda or settings reader quite so much (or at all). Let's chat about it?

src/main/scala/com/gu/kindlegen/LocalRun.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  override def getAwsRequestId: String = throw new UnsupportedOperationException
+  override def getLogGroupName: String = throw new UnsupportedOperationException
+  override def getLogger: LambdaLogger = throw new UnsupportedOperationException
+  override def getLogStreamName: String = throw new UnsupportedOperationException
+  override def getFunctionVersion: String = throw new UnsupportedOperationException
+  override def getInvokedFunctionArn: String = throw new UnsupportedOperationException
+  override def getIdentity: CognitoIdentity = throw new UnsupportedOperationException
+  override def getClientContext: ClientContext = throw new UnsupportedOperationException
+}
```
NathanielBennett, line 28: :+1: Good implementation. Provides you with what you need with minimal clutter. Ensures you bail if anything goes wrong

src/main/scala/com/gu/kindlegen/app/Lambda.scala
```diff
@@ -201,10 +203,17 @@ class Lambda(settings: Settings, date: LocalDate) extends Logging {
...
-          .failed.foreach(logger.warn("Publishing the zip archive failed!", _))
+        logger.info("Zipping publications now")
+
+        publisher.zipPublications().onComplete {
+          case Success(_) => {
+            logger.info("Publishing the zip archive succeeded! Deleting publications now")
+            publisher.close()
+            logger.info("All publications deleted")
+          }
+          case Failure(e) => logger.warn("Publishing the zip archive failed!", e)
```
regiskuckaertz, line 214: Surely The publisher should be closed here as well? Otherwise you risk leaking resources.

src/main/resources/reference.conf
```diff
@@ -9,7 +9,7 @@ books {
 
 content-api {
   key = "Get a key from https://bonobo.capi.gutools.co.uk"
-  url = "https://content.guardianapis.com"
+  url = "https://preview.content.guardianapis.com"
```
regiskuckaertz, line 12: 😲 this public?

regiskuckaertz (commented):
Hello 👋

PR #57, opened 2021-06-21
Updated to define the forthcoming Saturday print product.

blishen:
Adds the mappings for the new book and booksection tags for the Saturday magazine.

Specifically:

Saturday (book): https://tagmanager.gutools.co.uk/tag/90568
front (booksection):  https://tagmanager.gutools.co.uk/tag/90569
features (booksection): https://tagmanager.gutools.co.uk/tag/90570
culture (booksection): https://tagmanager.gutools.co.uk/tag/90571
lifestyle (booksection): https://tagmanager.gutools.co.uk/tag/90572

## What does this change?
When new print products are created we need to tell the Kindle Gen process two things:

Which book and booksection tags it should interrogate for content.
Where in the running order the content delivered by these tags should be shown.

## How to test
Tag some dummy content to meet the "print-sent" condition, and check it is included in a trial run in the correct position.
For completeness we should include at least one piece of content for each booksection tag.

## How can we measure success?
General delight from our readers that we've not neglected to include the new Saturday magazine content.

## Have we considered potential risks?
What, you mean apart from letting a product manager manager make PRs? We will need to make similar changes to the kindle web viewer, and update the Daily Edition prefills template with a PR on the fronts tool.


JustinPinner (approved):
LGTM 👍

Fweddi (commented):
I'm assuming the expected output of this is that articles marked up for the saturday edition will be generated as e.g. `42_saturday_2021_jun_22_article-name-here.nitf.xml`, and there will also be a general `theguardian_mainsection_saturday.rss.xml` index?

Happy to approve, but we can also test this in advance of merging by running locally or on CODE and checking S3.

